PR #44, opened 2019-01-24
On 1406 display standard name on standard confirmation page

gdbarnes:


src/SFA.DAS.ApplyService.Web/Views/Application/Submitted.cshtml
```diff
@@ -11,8 +11,19 @@
             <div class="govuk-panel govuk-panel--confirmation">
                 <h1 class="govuk-panel__title">Application submitted</h1>
                 <div class="govuk-panel__body">
-                    Reference number
-                    <br /><strong>@Model.ReferenceNumber</strong>
+                    <div class="govuk-!-margin-bottom-4">
+                        <span class="govuk-!-font-size-24">Standard applied for</span>
+                        <br />
+                        @if (!string.IsNullOrEmpty(Model.StandardName))
```
MarkFCain: Should this 'if' wrap the containing DIV?  this is a check the standard name is actually present

gdbarnes: Yes. Yes it should! 👍

PR #45, opened 2019-01-25
File upload improvements

davidgouge:
During apply process, if applicant goes back into a page with a file upload question on it, they now have the ability to download the uploaded file.  They can also delete the file.


src/SFA.DAS.ApplyService.Application/Apply/DeleteFile/DeleteFileHandler.cs
```diff
@@ -0,0 +1,77 @@
...
+//                poa.Answers.ForEach(a =>
+//                {
+//                    if (a.QuestionId == request.QuestionId)
+//                    {
+//                        poa.Answers.Remove(a);
+//                    }
+//                });
+//            });
+//            
+//
```
MarkFCain: What's all this commenting out? 👎

PR #48, opened 2019-01-25
On 1203 open closed applications

martin-hornsby:


src/SFA.DAS.ApplyService.Data/ApplyRepository.cs
```diff
@@ -382,6 +395,96 @@ ELSE @sectionStatusInProgress
...
+                        {
+                            applicationStatusApproved = ApplicationStatus.Approved,
+                            applicationStatusRejected = ApplicationStatus.Rejected,
+                            sequenceStatusApproved = ApplicationSequenceStatus.Approved,
+                            sequenceStatusRejected = ApplicationSequenceStatus.Rejected,
+                        })).ToList();
+            }
+        }
+
+
```
MarkFCain: should these big calls be in a sproc?   Just wondered.  Might be worth raising a tech debt story to do that?

martin-hornsby: If they get any bigger then I think we should. Currently it's a whole bunch of case statements based on sequence that is making it look huge

PR #60, opened 2019-01-29
On 1432 update declarations again

gdbarnes:
This includes a fix fro @MarkFCain to stop the title changing on the MultipleQuestions question type when a form is submitted. Example question is PageId: 20.1

### DIFF
--- a/src/SFA.DAS.ApplyService.Application/Apply/UpdatePageAnswers/UpdatePageAnswersHandler.cs
+++ b/src/SFA.DAS.ApplyService.Application/Apply/UpdatePageAnswers/UpdatePageAnswersHandler.cs
@@ -12,296 +12,296 @@ using SFA.DAS.ApplyService.Domain.Entities;
 
 namespace SFA.DAS.ApplyService.Application.Apply.UpdatePageAnswers
 {
-    public class UpdatePageAnswersHandler : IRequestHandler<UpdatePageAnswersRequest, UpdatePageAnswersResult>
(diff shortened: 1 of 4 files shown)


src/SFA.DAS.ApplyService.Application/Apply/UpdatePageAnswers/UpdatePageAnswersHandler.cs
```diff
@@ -12,296 +12,296 @@
...
+    //                    }
+    //                }
+    //                else
+    //                {
+    //                    workflow.Single(w => w.SequenceId == nextSequence.NextSequenceId).Active = true;
+    //                }
+    //            }
+    //        }
+  }
 }
```
davidgouge: I take it there's a small change in here somewhere? 🤔

davidgouge: @MarkFCain ?

MarkFCain: Greg did spot a name logic thing that was all about using the display type for logic which we resolved with a single line addition to reinject the displaytype on the viewmodel if there's a validation issue.  Not sure why it's gone change crazy.  It was line 43 btw.

davidgouge: 43, that's what I was after. Ta.

PR #66, opened 2019-01-31
On 1400 financial due date

davidgouge:


src/SFA.DAS.ApplyService.Application/Apply/Financial/UpdateGradeHandler.cs
```diff
@@ -4,33 +4,50 @@
...
+            return org;
+        }
+
+        private async Task<Organisation> UpdateApplyOrganisation(UpdateGradeRequest request)
+        {
+            var org = await _organisationRepository.GetOrganisationByApplicationId(request.ApplicationId);
+            org.OrganisationDetails.FinancialDueDate = request.UpdatedGrade.FinancialDueDate;
+            org.OrganisationDetails.FinancialExempt =
+                request.UpdatedGrade.SelectedGrade == FinancialApplicationSelectedGrade.Exempt;
+            await _organisationRepository.UpdateOrganisation(org, Guid.NewGuid());
```
MarkFCain: This looks wrong.  I think we pass a userid in as the guid here, so doesn't that need the userid?

src/SFA.DAS.ApplyService.InternalApi/Infrastructure/AssessorServiceApiClient.cs
```diff
@@ -107,6 +110,11 @@ private string GetToken()
 
             return result.AccessToken;
         }
+
+        public async Task UpdateOrganisation(Organisation organisation)
+        {
+            //var org = GetOrganisationByEmail()
```
MarkFCain: looks moot

davidgouge:
Nice spots @MarkFCain 😁 👍

PR #76, opened 2019-02-04
Add analytics tagging as on sfa.das.assessorservice.web

gdbarnes:


src/SFA.DAS.ApplyService.Web/Views/Shared/_Layout.cshtml
```diff
@@ -35,10 +39,16 @@
 </head>
 
 <body class="govuk-template__body ">
+
 <script>
     document.body.className = ((document.body.className) ? document.body.className + ' js-enabled' : 'js-enabled');
 </script>
 
+<environment exclude="Development">
```
MarkFCain: Is it only test we want to exclude? I don't know the answer, but thought I'd ask

gdbarnes: I guess in this case we'd want to exclude both the Test and Development environments.

Do you know if we're set up to do something like:
`<environment include="Production">`?

It looks like the docs for this kind of thing are here:
https://docs.microsoft.com/en-us/aspnet/core/fundamentals/environments?view=aspnetcore-2.2

MarkFCain: Not a clue

PR #84, opened 2019-02-06
On 1457 skip sequence one if finances up to date

davidgouge:
This PR is merging to 1400 as it's an extension of that work and relies on it.   It also includes some tidying / removing / warnings fixes, so apols.


src/SFA.DAS.ApplyService.Application/Apply/StartApplicationHandler.cs
```diff
@@ -45,8 +46,60 @@ public async Task<Unit> Handle(StartApplicationRequest request, CancellationToke
...
+
             await _applyRepository.UpdateSections(sections);
+            await _applyRepository.UpdateSequences(sequences);
+            
             return Unit.Value;
         }
+
+        private void DisableSequencesAndSectionsAsAppropriate(Organisation org, List<ApplicationSequence> sequences, List<ApplicationSection> sections)
+        {
+            // IF IsEPAOApproved = true;
```
MarkFCain: remove this

davidgouge: /giphy salute
😆

PR #88, opened 2019-02-11
On 1295 companies house api

martin-hornsby:


src/SFA.DAS.ApplyService.InternalApi/Infrastructure/CompaniesHouseApiClient.cs
```diff
@@ -0,0 +1,135 @@
...
+        {
+            _logger.LogInformation($"Searching Companies House - People With Significant Control. Company Number: {companyNumber}");
+            var apiResponse = await Get<PersonWithSignificantControlList>($"/company/{companyNumber}/persons-with-significant-control?items_per_page=100");
+
+            var items = activeOnly ? apiResponse.items?.Where(i => i.ceased_on is null) : apiResponse.items;
+            return Mapper.Map<IEnumerable<PersonWithSignificantControl>, IEnumerable<Types.CompaniesHouse.PersonWithSignificantControl>>(items);
+        }
+
+        private async Task<IEnumerable<dynamic>> GetCharges(string companyNumber)
+        {
```
martin-hornsby: this is not being used at the moment but is just a reminder that a mapper would need to be created.

src/SFA.DAS.ApplyService.InternalApi/Infrastructure/CompaniesHouseApiClient.cs
```diff
@@ -0,0 +1,135 @@
...
+            _logger.LogInformation($"Searching Companies House - Charges. Company Number: {companyNumber}");
+            var apiResponse = await Get<ChargeList>($"/company/{companyNumber}/charges");
+            return apiResponse.items;
+            //return Mapper.Map<IEnumerable<Charge>, IEnumerable<Types.CompaniesHouse.Charge>> (apiResponse.items);
+        }
+
+        private async Task<dynamic> GetInsolvencyDetails(string companyNumber)
+        {
+            _logger.LogInformation($"Searching Companies House - Insolvency. Company Number: {companyNumber}");
```
martin-hornsby: this is not being used at the moment but is just a reminder that a mapper would need to be created.

PR #100, opened 2019-02-21
ON-1456 - Exempt Questions

martin-hornsby:


src/SFA.DAS.ApplyService.Application/Apply/StartApplicationHandler.cs
```diff
@@ -38,6 +38,14 @@ public async Task<Unit> Handle(StartApplicationRequest request, CancellationToke
 
             foreach (var applicationSection in sections)
             {
+                var pagesToMakeNotRequired = applicationSection.QnAData.Pages.Where(p => p.NotRequiredOrgTypes.Contains(org.OrganisationType));
+
```
davidgouge: To be on the safe side, check here that NotRequiredOrgTypes is not null.  At least then you'd only have to add the property on those Pages in the workflow that _are_ affected. 😁

martin-hornsby: Originally I was null checking that then saw in the workflow that other IEnumerable types didn't have this check (but an empty array in the JSON). Tbh it's probably better we have the check there :)

PR #105, opened 2019-03-01
On 1564 charity commission api

martin-hornsby:


src/SFA.DAS.ApplyService.InternalApi/Connected Services/CharityCommissionService/Entities/Charity.cs
```diff
@@ -0,0 +1,61 @@
...
+                return dates?.Min();
+            }
+        }
+
+        [XmlIgnore]
+        public string[] NatureOfBusiness
+        {
+            get
+            {
+                return Classification?.What;
```
MarkFCain: that's a hell of a field name

src/SFA.DAS.ApplyService.InternalApi/Connected Services/CharityCommissionService/Entities/LatestFiling.cs
```diff
@@ -0,0 +1,41 @@
...
+        [XmlIgnore]
+        public DateTime? AnnualReturnPeriodDateTime
+        {
+            get
+            {
+                try
+                {
+                    return XmlConvert.ToDateTime(this.annualReturnPeriodField, @"dd/MM/yyyy");
+                }
+                catch
```
MarkFCain: I'm no fan of using catch from programmatic reasons.  If this is the only alternative, then that ok, but if there is a better way to check validity, can you do that instead?

src/SFA.DAS.ApplyService.InternalApi/Connected Services/CharityCommissionService/Entities/Charity.cs
martin-hornsby: sadly, this comes from the SOAP Endpoint

martin-hornsby:
Replaced XmlConvert.ToDateTime with DateTime.TryParseExact :)

PR #113, opened 2019-03-19
On 1623 char and word limits

alandburns:


src/SFA.DAS.ApplyService.Application/Apply/Validation/MaxWordValueValidator.cs
```diff
@@ -0,0 +1,37 @@
...
+            int wordCount = 0, index = 0;
+
+            while (index < text.Length)
+            {
+                while (index < text.Length && !char.IsWhiteSpace(text[index]))
+                    index++;
+                wordCount++;
+                while (index < text.Length && char.IsWhiteSpace(text[index]))
+                    index++;
+            }
```
davidgouge: How about `int wordCount = text.Split(" ", StringSplitOptions.RemoveEmptyEntries).Length;` ?? 🤔

PR #127, opened 2019-03-28
On 1173 secondary app submissions

martin-hornsby:


src/SFA.DAS.ApplyService.Data/ApplyRepository.cs
```diff
@@ -431,11 +465,38 @@ public async Task UpdateApplicationStatus(Guid applicationId, string status)
...
+                                                            SET  IsActive = 0, Status = @rejectedStatus
+                                                            WHERE  ApplicationSequences.ApplicationId = @applicationId;
+
+                                                            UPDATE Applications
+                                                            SET  ApplicationStatus = @rejectedSequenceStatus, DeletedAt = GETUTCDATE(), DeletedBy = 'System'
+                                                            WHERE  Applications.Id = @applicationId;",
+                                                            new { applicationId = app.Id, rejectedStatus = ApplicationStatus.Rejected, rejectedSequenceStatus = ApplicationSequenceStatus.Rejected });
+                        }
+                    }
+                }
```
davidgouge: Seems to be quite a lot of business logic going on in this repository. Shouldn't this be in .Application and therefore testable? 😁

src/SFA.DAS.ApplyService.Web/Controllers/ApplicationController.cs
```diff
@@ -565,9 +570,16 @@ public async Task<IActionResult> Submit(Guid applicationId, int sequenceId)
...
+
+            if (await _apiClient.Submit(applicationId, sequenceId, User.GetUserId(), User.GetEmail()))
+            {
+                return RedirectToAction("Submitted", new { applicationId });
+            }
+            else
+            {
+                // unable to submit
+                return RedirectToAction("NotSubmitted", new { applicationId });
+            }
```
davidgouge: Are there any tests that can cover these new rules in this controller?

src/SFA.DAS.ApplyService.Data/ApplyRepository.cs
martin-hornsby: done

src/SFA.DAS.ApplyService.Web/Controllers/ApplicationController.cs
martin-hornsby: Added unit tests to the handler (which pass this on the result to the controller)

PR #135, opened 2019-04-09
ON-1829 New file upload

davidgouge:
Uses a JS front end to enable drag and drog, multiple uploads, chunking upload etc.
Saves each chunk to storage.  Once all chunks accounted for reassembles them into the original file, encrypts it and stores it.


alandburns:
this is history

PR #141, opened 2019-04-15
On 1737 show user org in header

gdbarnes:


src/SFA.DAS.ApplyService.Web/Controllers/UsersController.cs
```diff
@@ -100,19 +102,28 @@ public async Task<IActionResult> PostSignIn()
         {
             var user = await _usersApiClient.GetUserBySignInId(User.GetSignInId());
            
-            if (user == null)
+            if (user is null)
```
davidgouge: fancy. 😁  Clearer than `== null` though?

martin-hornsby: is has been the preferred way since C# 7.0

== could be be overriden and also has issues when matching nulls as it tends to require intialized lhs & rhs objects

https://docs.microsoft.com/en-us/dotnet/csharp/whats-new/csharp-7#pattern-matching

PR #147, opened 2019-04-18
Fixed incorrect log error message

martin-hornsby:


martin-hornsby:
@reachash @umakanth-g - to test this one you'll need to check emails are being sent on application submissions and there's an log entry in Kibana

PR #153, opened 2019-04-26
On 1773 multiple answer questions

davidgouge:


src/SFA.DAS.ApplyService.Web/Views/Users/CreateAccount.cshtml
```diff
@@ -61,8 +83,17 @@
...
-                    <input class="govuk-input" id="Email" name="Email" type="text" asp-for="Email"
-                        aria-describedby="email-hint" sfa-validation-for="FamilyName" sfa-validationerror-class="govuk-input--error" />
+                    <input
+                        class="govuk-input"
+                        id="Email"
+                        name="Email"
+                        type="text"
+                        asp-for="Email"
+                        aria-describedby="email-hint"
+                        sfa-validation-for="FamilyName"
```
ChrisJWoodcock: This should Email - I know it was pre-existing code.

src/SFA.DAS.ApplyService.Web/Views/Users/CreateAccount.cshtml
```diff
@@ -61,8 +83,17 @@
...
+                        sfa-validationerror-class="govuk-input--error"
+                        maxlength="320"
```
ChrisJWoodcock: There doesn't seem to be anything specific in the ticket that suggests max length would be changed - or what the correct value should be (I'm guessing that this is the Email field in [Apply].[Contacts] - which is 256 or could eventually end up there so would be best to be the same max length)

src/SFA.DAS.ApplyService.Web/Views/Users/CreateAccount.cshtml
```diff
@@ -47,7 +58,18 @@
...
+                        class="govuk-input govuk-!-width-two-thirds"
+                        id="FamilyName"
+                        name="FamilyName"
+                        type="text"
+                        asp-for="FamilyName"
+                        sfa-validation-for="FamilyName"
+                        sfa-validationerror-class="govuk-input--error"
+                        autocomplete="family-name"
+                        spellcheck="false"
+                        maxlength="10000"
```
ChrisJWoodcock: Perhaps this should be the same length as most Family names in the Database e.g. 250

src/SFA.DAS.ApplyService.Web/Views/Users/CreateAccount.cshtml
```diff
@@ -36,7 +36,18 @@
...
+                        class="govuk-input govuk-!-width-two-thirds"
+                        id="GivenName"
+                        name="GivenName"
+                        type="text"
+                        asp-for="GivenName"
+                        sfa-validation-for="GivenName"
+                        sfa-validationerror-class="govuk-input--error"
+                        autocomplete="given-name"
+                        spellcheck="false"
+                        maxlength="10000"
```
ChrisJWoodcock: Perhaps this should be the same length as most Given names in the Database e.g. 250

src/SFA.DAS.ApplyService.Web/Views/Application/Standard/FindStandard.cshtml
```diff
@@ -40,7 +40,7 @@
                         <span id="search-string-error" class="govuk-error-message">Enter 2 or more characters to search</span>
                     }
 
-                    <input class="govuk-input govuk-input--width-20 ui-autocomplete-input @(!ViewData.ModelState.IsValid || showErrors ? "govuk-input--error" : "")" id="standard-name" name="StandardToFind" asp-for="@Model.StandardToFind"type="text" value="" autocomplete="off" maxlength="500">
+                    <input class="govuk-input govuk-input--width-20 ui-autocomplete-input @(!ViewData.ModelState.IsValid || showErrors ? "govuk-input--error" : "")" id="standard-name" name="StandardToFind" asp-for="@Model.StandardToFind"type="text" value="" autocomplete="off" maxlength="10000">
```
ChrisJWoodcock: Why is this increase by so much - is this the same as StandardCollation (Title) which is 500 character is the Assessor database.

src/SFA.DAS.ApplyService.Web/Views/Application/Standard/FindStandard.cshtml
```diff
@@ -40,7 +40,7 @@
                         <span id="search-string-error" class="govuk-error-message">Enter 2 or more characters to search</span>
                     }
 
-                    <input class="govuk-input govuk-input--width-20 ui-autocomplete-input @(!ViewData.ModelState.IsValid || showErrors ? "govuk-input--error" : "")" id="standard-name" name="StandardToFind" asp-for="@Model.StandardToFind"type="text" value="" autocomplete="off">
+                    <input class="govuk-input govuk-input--width-20 ui-autocomplete-input @(!ViewData.ModelState.IsValid || showErrors ? "govuk-input--error" : "")" id="standard-name" name="StandardToFind" asp-for="@Model.StandardToFind"type="text" value="" autocomplete="off" maxlength="500">
```
ChrisJWoodcock: In a subsequent change a similar field is increase from 500 to 10000 - this is given a new 500 limit though - inconsistent?

gdbarnes: This is the same file, so is outdated.

src/SFA.DAS.ApplyService.Web/Views/Users/CreateAccount.cshtml
ChrisJWoodcock:
```suggestion
                        sfa-validation-for="Email"
```

src/SFA.DAS.ApplyService.Web/Views/Users/CreateAccount.cshtml
ChrisJWoodcock: The reason for increase was to match 10000 as that is the character limit in the WAF;

PR #156, opened 2019-04-26
On 1814 premable confirm org type

martin-hornsby:


src/SFA.DAS.ApplyService.Web/Views/OrganisationSearch/Results.cshtml
```diff
@@ -55,7 +55,7 @@
                         {
                             <tr class="govuk-table__row govuk-table__row--no-top-padding-on-first-child">
                                 <td class="govuk-table__cell govuk-!-padding-bottom-4">
-                                    <form method="post" asp-action="@(result.OrganisationType != null ? "Confirm" : "Type")" asp-controller="OrganisationSearch" novalidate>
+                                    <form method="post" asp-action="@(result.RoEPAOApproved ? "Confirm" : "Type")" asp-controller="OrganisationSearch" novalidate>
```
davidgouge: Can this c# be pushed into the VM?

martin-hornsby: Will be quite difficult to do as this VM is shared across the Preamble journey

davidgouge: Surely a new OrganisationSearchAction property with that code in wouldn't hurt anywhere else?

martin-hornsby: So it turns out it is an internal API type rather than a VM. @davidgouge are you happy to approve if we resolve this issue another time?

PR #157, opened 2019-04-29
Revert "Added a redirect to access denied if contacts.status is Deleted"

alandburns:
Reverts SkillsFundingAgency/das-apply-service#155


alandburns:
Not needed

PR #174, opened 2019-05-14
ON-1732 - Updated Apply EmailTemplate Ids used

martin-hornsby:


martin-hornsby:
I don't see these templates being used in Apply, however they are in the database so for completeness I've updated them.

alandburns:
This is now only changes to Email templates - see https://skillsfundingagency.atlassian.net/browse/ON-1637

PR #183, opened 2019-05-22
Apr 446 determine verification required

David-B-Read:


src/SFA.DAS.ApplyService.Domain/Ukrlp/ProviderDetails.cs
```diff
@@ -12,6 +13,29 @@ public class ProviderDetails
         public DateTime? VerificationDate { get; set; }
         public List<ProviderAlias> ProviderAliases { get; set; }
         public List<VerificationDetails> VerificationDetails { get; set; }
+
+        public bool VerifiedByCompaniesHouse
+        {
+            get
+            {
+                var companiesHouseVerification = VerificationDetails.FirstOrDefault(x =>
```
MarkFCain: I could be wrong but I think you could just change this to something like
return VerificationDetails.Any(x => ...)

PR #190, opened 2019-06-04
Apr 485 store organisation details

David-B-Read:


src/SFA.DAS.ApplyService.Domain/Ukrlp/ProviderDetails.cs
```diff
@@ -32,6 +32,11 @@ public bool VerifiedbyCharityCommission
             }
         }
 
+        public ProviderContact PrimaryContactDetails
+        {
+            get { return ContactDetails.FirstOrDefault(x => x.ContactType == "L"); }
```
MarkFCain: minor, minor point.  Maybe define "L" as a named variable describing what it is? It's minor as it is used in one place, where it's use is evident, so I'll leave that with you.

src/SFA.DAS.ApplyService.Application.UnitTests/Handlers/StartApplicationHandlerTests/StartApplicationHandlerTestsBase.cs
```diff
@@ -26,12 +27,13 @@ public void Setup()
             var latestWorkflowId = Guid.NewGuid();
             ApplyingOrganisationId = Guid.NewGuid();
             ApplicationId = Guid.NewGuid();
+            ApplicationType = "EPAO";
```
MarkFCain: change this to ApplicationTypes.EndpointAssessor maybe?

PR #191, opened 2019-06-05
On 1859 invite user to organisation

davidgouge:
Support for Assessor's Invite User to the Organisation work.

### DIFF
--- a/src/SFA.DAS.ApplyService.Application/Users/CreateNewContact/CreateNewContactRequest.cs
+++ b/src/SFA.DAS.ApplyService.Application/Users/CreateNewContact/CreateNewContactRequest.cs
@@ -1,0 +1,17 @@
+using System;
+using MediatR;
+
+namespace SFA.DAS.ApplyService.Application.Users.CreateNewContact
+{
+    public class CreateNewContactRequest : IRequest
+    {
+        public Guid Id { get; set; }
+        public Guid OrganisationId { get; set; }
+        public string Email { get; set; }
+        public string GivenNames { get; set; }
+        public string FamilyName { get; set; }
+        public string DisplayName { get; set; }
+        public string Status { get; set; }
+        public string SignInType { get; set; }
+    }
+}
--- a/src/SFA.DAS.ApplyService.InternalApi.Types/SignInCallback.cs
+++ b/src/SFA.DAS.ApplyService.InternalApi.Types/SignInCallback.cs
@@ -2,7 +2,7 @@ using System;
 
 namespace SFA.DAS.ApplyService.InternalApi.Types
 {
-    public class DfeSignInCallback
+    public class SignInCallback
     {
         public string Sub { get; set; }  
         public string SourceId { get; set; }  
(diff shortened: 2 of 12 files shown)


src/SFA.DAS.ApplyService.Application/Users/CreateNewContact/CreateNewContactRequest.cs
```diff
@@ -0,0 +1,17 @@
...
+namespace SFA.DAS.ApplyService.Application.Users.CreateNewContact
+{
+    public class CreateNewContactRequest : IRequest
+    {
+        public Guid Id { get; set; }
+        public Guid OrganisationId { get; set; }
+        public string Email { get; set; }
+        public string GivenNames { get; set; }
+        public string FamilyName { get; set; }
+        public string DisplayName { get; set; }
```
ChrisJWoodcock: Is this used?

src/SFA.DAS.ApplyService.InternalApi.Types/SignInCallback.cs
```diff
@@ -2,7 +2,7 @@
 
 namespace SFA.DAS.ApplyService.InternalApi.Types
 {
-    public class DfeSignInCallback
+    public class SignInCallback
```
ChrisJWoodcock: I was previously thinking that Dfe was being left all over the code in case we migrated back to the Dfe service?

davidgouge: LOL. If we do migrate back, then SignInCallback is still valid.

davidgouge: And after all the work put into ASLogin, we're not going back to DfE. 😉

src/SFA.DAS.ApplyService.Application/Users/CreateNewContact/CreateNewContactRequest.cs
davidgouge: Yep

src/SFA.DAS.ApplyService.InternalApi.Types/SignInCallback.cs
martin-hornsby: I like it being generic. I guess if we have multiple SignIn systems then make it abstract at that point

src/SFA.DAS.ApplyService.Application/Users/CreateNewContact/CreateNewContactRequest.cs
martin-hornsby: Whereabouts? I can't see it being used in the Handler

src/SFA.DAS.ApplyService.InternalApi.Types/SignInCallback.cs
ChrisJWoodcock: I think a Tech Debt needs raising then to remove all mentions of Dfe Sign in - some of the log messages you added mention it in the new code as well?

PR #202, opened 2019-06-20
ON-1482 postcode lookup

gdbarnes:


src/SFA.DAS.ApplyService.Application/Apply/Validation/DateNotInFutureValidator.cs
```diff
@@ -6,14 +6,13 @@
 
 namespace SFA.DAS.ApplyService.Application.Apply.Validation
 {
-    public class DateNotInFutureValidator : IValidator
+    public class DateNotInFutureValidator : Validator
```
martin-hornsby: As you're inheriting from an abstract class you really ought to be making these things a sealed class.
Not a biggie but worth keeping in mind

src/SFA.DAS.ApplyService.Web/wwwroot/javascripts/postcode-lookup-service.js
```diff
@@ -215,4 +227,11 @@
         $(".address-manual-input").removeClass("js-hidden");
         $(".address-manual-input-focus").focus();
     }
+
+    function searchAgain() {
```
ChrisJWoodcock: Is this supposed to be empty?

src/SFA.DAS.ApplyService.Application/Apply/Validation/DateNotInFutureValidator.cs
ChrisJWoodcock: I don't think this is a thing unless a article can be found to support sealing inherited abstract classes - I can't find anything on Google.

src/SFA.DAS.ApplyService.InternalApi/Startup.cs
```diff
@@ -172,56 +175,49 @@ public void Configure(IApplicationBuilder app, IHostingEnvironment env)
...
+            services.AddTransient<IValidator, DateNotInFutureValidator>();
+            services.AddTransient<IValidator, DateValidator>();
+            services.AddTransient<IValidator, EmailAddressIsValidValidator>();
+            services.AddTransient<IValidator, MaxLengthValidator>();
+            services.AddTransient<IValidator, MaxWordCountValidator>();
+            services.AddTransient<IValidator, NullValidator>();
+            services.AddTransient<IValidator, RegexValidator>();
+            services.AddTransient<IValidator, RegisteredCharityNumberValidator>();
+            services.AddTransient<IValidator, RequiredValidator>();
+            services.AddTransient<IValidator, SimpleRadioNotNullValidator>();
```
davidgouge: Ideally, all these IValidator wireups need replacing with some sort of "AddAllTypesThatImplement<IValidator>" call as the idea is that we can add IValidators without worrying about wiring up.  But as this will all be gone soon with the implementation of Qna Api, I'm just leaving this as a comment.

PR #205, opened 2019-06-21
Apr 513 change application flow

David-B-Read:
Moved select UKPRN to first page of preamble and changed select application route from a controller action to a question in the workflow


src/SFA.DAS.ApplyService.Domain/Ukrlp/ProviderDetails.cs
```diff
@@ -68,6 +68,34 @@ public class ContactAddress
         public string Address4 { get; set; }
         public string Town { get; set; }
         public string PostCode { get; set; }
+
```
MarkFCain: not for here, but we may need to move this kind of thing closer to the view.  Let's discuss with Greg some time...

David-B-Read: Yeah I'll move the format function into the VM

PR #219, opened 2019-07-15
 ON-2068 Allow Existing EPAO to apply for multiple standards

arshedmust:
As discussed with Deen the second scenario in this story also takes you to the list of application screen, and from there a new application can be started and then the current follow will take care of financials.

### DIFF
--- a/src/SFA.DAS.ApplyService.Web/Controllers/ApplicationController.cs
+++ b/src/SFA.DAS.ApplyService.Web/Controllers/ApplicationController.cs
@@ -48,7 +48,7 @@ namespace SFA.DAS.ApplyService.Web.Controllers
                 return RedirectToAction("PostSignIn", "Users");
 
             _logger.LogInformation($"Got LoggedInUser from Session: {user}");
-            
+
             var applyUser = await _usersApiClient.GetUserBySignInId((await _userService.GetSignInId()).ToString());
             var userId = applyUser?.Id ?? Guid.Empty;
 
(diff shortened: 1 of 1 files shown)


src/SFA.DAS.ApplyService.Web/Controllers/ApplicationController.cs
```diff
@@ -58,48 +58,44 @@ public async Task<IActionResult> Applications()
...
-                if (org == null)
-                {
-                    if (await _userService.AssociateOrgFromClaimWithUser())
-                        return await StartApplication(userId);
+                if (org != null)
+                    return View("~/Views/Application/Declaration.cshtml");
 
-                    return RedirectToAction("Index", "OrganisationSearch");
-                }
+                if (await _userService.AssociateOrgFromClaimWithUser())
```
ChrisJWoodcock: Has the logic here changed? If it hasn't it's not necessary to change the code?

src/SFA.DAS.ApplyService.Web/Controllers/ApplicationController.cs
```diff
@@ -58,48 +58,44 @@ public async Task<IActionResult> Applications()
...
+
             if (applications.Count() > 1)
-            {
                 return View(applications);
-            }
 
             var application = applications.First();
 
-            if (application.ApplicationStatus == ApplicationStatus.FeedbackAdded)
+            switch (application.ApplicationStatus)
```
ChrisJWoodcock: Has the logic changed? - if it hasn't I wouldn't change the code - as you 'own' it now and questions such as 'When using First() you aren't guaranteed an order?' become your's to answer; when they weren't before...

src/SFA.DAS.ApplyService.Web/Controllers/ApplicationController.cs
```diff
@@ -58,48 +58,44 @@ public async Task<IActionResult> Applications()
...
-            if (application.ApplicationStatus == ApplicationStatus.Approved)
-            {
-                return View(applications);
+                case ApplicationStatus.FeedbackAdded:
+                    return View("~/Views/Application/FeedbackIntro.cshtml", application.Id);
+                case ApplicationStatus.Rejected:
+                case ApplicationStatus.Approved:
+                    return View(applications);
+                default:
+                    return RedirectToAction("SequenceSignPost", new {applicationId = application.Id});
```
ChrisJWoodcock: Does this code have test coverage?

src/SFA.DAS.ApplyService.Web/Controllers/ApplicationController.cs
```diff
@@ -58,48 +58,44 @@ public async Task<IActionResult> Applications()
 
             if (!applications.Any())
             {
+                //ON-2068 Registered org  with no application created via digital service then
```
ChrisJWoodcock: I don't think we should be adding the ON number to the comments as this should be traceable to the ticket which should have enough details to find out why this is being done - also the title of your pull request does not match this ticket number

ChrisJWoodcock: For my knowledge does Created by the digital service align with org.RoEPAOApproved - if you create this some other way will it not have that flag?

src/SFA.DAS.ApplyService.Web/Controllers/ApplicationController.cs
arshedmust: There where more then 3 ifs in a row so clean the code using switch statement

src/SFA.DAS.ApplyService.Web/Controllers/ApplicationController.cs
martin-hornsby: Probably not, although I can tell that the logic for this part hasn't changed. Making it a switch is tidier

src/SFA.DAS.ApplyService.Web/Controllers/ApplicationController.cs
```diff
@@ -58,48 +58,44 @@ public async Task<IActionResult> Applications()
...
-                {
-                      return await StartApplication(userId);
-                }
+                    return View(applications);
 
-                if (org == null)
-                {
-                    if (await _userService.AssociateOrgFromClaimWithUser())
-                        return await StartApplication(userId);
+                if (org != null)
```
martin-hornsby: Could you tidy this up by doing something like this? 
`                if (org != null)
                {
                    return org.RoEPAOApproved ? return View(applications) : return View("~/Views/Application/Declaration.cshtml");
                }`

src/SFA.DAS.ApplyService.Web/Controllers/ApplicationController.cs
arshedmust: As far as I know the all registered orgs will have this flag set. So if there is no applications for a registered org then the org is assummed to be created out side of apply.

src/SFA.DAS.ApplyService.Web/Controllers/ApplicationController.cs
arshedmust: I have updated the code with:
 if (org != null) return org.RoEPAOApproved ?  View(applications) : View("~/Views/Application/Declaration.cshtml");

src/SFA.DAS.ApplyService.Web/Controllers/ApplicationController.cs
arshedmust: Tided up the code the logic is the same

src/SFA.DAS.ApplyService.Web/Controllers/ApplicationController.cs
arshedmust: By putting ON numbers by fixes helps future developers to trace why this change was made especially the branch is merged with master which may have many other changes.

PR #222, opened 2019-07-19
Add new YAML build definition

EwanNoble:


EwanNoble:
@martin-hornsby @davidgouge Looks like the Web.IntegrationTests have been missing from the Apply solution and were therefore missed out of the move to 2.2. Have had to hack it apart to get it building. Would someone be able to do a proper pass over it and get it working again? 🙂

PR #227, opened 2019-07-23
ON-2117 - Refactoring & Error logging added for Blob Storage

martin-hornsby:


src/SFA.DAS.ApplyService.Web/Infrastructure/FileValidator.cs
```diff
@@ -26,27 +26,65 @@ public static bool FileValidationPassed(List<Answer> answers, Page page, List<Va
...
+        {
+            {"JPG",  new byte[] { 0xFF, 0xD8, 0xFF }},
+            {"JPEG", new byte[] { 0xFF, 0xD8, 0xFF }},
+            {"TIF",  new byte[] { 0x49, 0x49, 0x2A, 0x00 }},
+            {"TIFF", new byte[] { 0x49, 0x49, 0x2A, 0x00 }},
+            {"GIF",  new byte[] { 0x47, 0x49, 0x46, 0x38 }},
+            {"BMP",  new byte[] { 0x42, 0x4D }},
+            {"ICO",  new byte[] { 0x00, 0x00, 0x01, 0x00 }},
+            {"PDF",  new byte[] { 0x25, 0x50, 0x44, 0x46 }}
+        };
```
davidgouge: I like this, but why are we now taking responsibility to make sure a file is actually a type of that file?  If the user uploads a .pdf, but it's not really a pdf, then that's their problem surely and they shouldn't be renaming extensions etc. Making this check and showing an error then pushes the problem back to us as we have to then explain why the file isn't uploading.

alandburns:
Superseded by QnA API

PR #233, opened 2019-07-30
turn off ShowPII

alandburns:


src/SFA.DAS.ApplyService.InternalApi/Startup.cs
```diff
@@ -56,7 +56,7 @@ public Startup(IConfiguration configuration, IHostingEnvironment env)
         public IServiceProvider ConfigureServices(IServiceCollection services)
         {
 
-            IdentityModelEventSource.ShowPII = true;
+            IdentityModelEventSource.ShowPII = false;
```
davidgouge: You can actually remove this line as it defaults to false.

alandburns: Yep. I know and thought about that.  Can do.

PR #235, opened 2019-08-01
ON-1975 Added DateOfBirth as an answer type to remove day from date of birth …

alandburns:
…questions.


alandburns:
According to comment in JIRA https://skillsfundingagency.atlassian.net/browse/ON-1975 this is built on top of ON-1482_postcode_lookup 

Hence https://github.com/SkillsFundingAgency/das-apply-service/pull/202 should be the base branch.

ChrisJWoodcock:
> According to comment in JIRA https://skillsfundingagency.atlassian.net/browse/ON-1975 this is built on top of ON-1482_postcode_lookup
> 
> Hence #202 should be the base branch.

I think the original intention was to base this on the multi-part json answers from the post-code lookup branch but I didn't do that in the end; however @davidgouge to merge that branch into this one; so now it is based on that branch so it should merge after that one; but is Ok to test now.

PR #242, opened 2019-08-15
APR-512 schema changes in separate branch so can be deployed to test envs

David-B-Read:


David-B-Read:
Abandoning this PR because it's incompatible with the new QnA database schema. Changes will be moved into a custom config section

PR #259, opened 2019-09-12
Apr 525 confirm officers pscs

David-B-Read:


David-B-Read:
> I assume that this will make use the Nuget version of QnA API Packages to display the Table values, that is currently coded explicitly in this branch.
> TabularData.cs
> _TabularData.cshtml

The QnA Api.Views project only has 4 question partial views, not the full set - should this be updated first?

David-B-Read:
This PR is now dependent on this QnA PR:
https://github.com/SkillsFundingAgency/das-qna-api/pull/33

PR #266, opened 2019-09-24
Apr 719 protecting your apprentices intro

MarkFCain:
NB: I want to add some coverage tests to this, but the core code is done, so don't let that stop you


src/RoatpWorkflow-QnA.sql
```diff
@@ -2295,3 +2295,295 @@ INSERT INTO [dbo].[WorkflowSequences]
            ,1)
 GO
 
+
+
+
+--- APR-719 adding sequence 4, section 1 interactions
+DECLARE @WorkflowId UNIQUEIDENTIFIER
```
David-B-Read: These variables are already defined on line 1 of this script

MarkFCain: ah I just reincluded them to be a distinct thing for the story, as previous section ended with the GO.   
I'll integrate them as one call now

src/SFA.DAS.ApplyService.Web/Infrastructure/Services/ProcessPageFlowService.cs
```diff
@@ -0,0 +1,70 @@
...
+                default:
+                    return null;
+            }
+        }
+
+
+        public async Task<int> GetApplicationProviderTypeId(Guid applicationId)
+        {
+            var providerTypeId = 1;
+            string pageId = RoatpWorkflowPageIds.YourOrganisationIntroductionMain;
```
David-B-Read: Is this needed?

src/SFA.DAS.ApplyService.Web/Infrastructure/Interfaces/IProcessPageFlowService.cs
```diff
@@ -0,0 +1,14 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+
+namespace SFA.DAS.ApplyService.Web.Infrastructure.Interfaces
+{
+    public interface IProcessPageFlowService
+    {
+        Task<string> GetIntroductionPageIdForSection(Guid applicationId, int sequenceId, int providerTypeId);
```
David-B-Read: This accepts a sequenceId, would it be more appropriate to name is GetIntroductionPageIdForSequence?

src/SFA.DAS.ApplyService.Web/Infrastructure/Services/ProcessPageFlowService.cs
```diff
@@ -0,0 +1,69 @@
...
+
+        public async Task<string> GetIntroductionPageIdForSection(Guid applicationId, int sequenceId,
+            int providerTypeId)
+        {
+            var sequenceDescription = _configuration.FirstOrDefault(x => x.Id == sequenceId);
+
+            if (sequenceDescription == null)
+                return null;
+
+            switch (providerTypeId)
```
David-B-Read: Rather than having this mapping from id to provider type name, would it be easier to reference the provider type id (1-3) in config?

src/SFA.DAS.ApplyService.Web/Infrastructure/Interfaces/IProcessPageFlowService.cs
MarkFCain: yep!

src/SFA.DAS.ApplyService.Web/Infrastructure/Services/ProcessPageFlowService.cs
MarkFCain: I was avoiding magic numbers by doing this, but I think the code smell cuts both ways.  I think I'll take your suggestion though, I can't think how to make this more elegant

src/SFA.DAS.ApplyService.Web.UnitTests/Services/ProcessPageFlowServiceTests.cs
```diff
@@ -0,0 +1,80 @@
...
+namespace SFA.DAS.ApplyService.Web.UnitTests.Services
+{
+    [TestFixture]
+    public class ProcessPageFlowServiceTests
+    {
+        private ProcessPageFlowService _service;
+        private Mock<IQnaApiClient> _qnaApiClient;
+        private Mock<IOptions<List<TaskListConfiguration>>> _configuration;
+
+        //IQnaApiClient qnaApiClient,
```
David-B-Read: Can these be removed?

PR #297, opened 2019-10-09
APR-700 attempt to retrieve QnA definitions from cache first

David-B-Read:


David-B-Read:
> Would it be worth implementing this as a generic cache aside pattern in a new function in ISessionService? E.g. https://stackoverflow.com/questions/44504740/what-is-a-good-way-to-implement-cache-aside-pattern

The ISessionService service is used by other parts of the Apply journey, not just this Api client. I think it is better structured to have the session service "dumb" and only responsible for storing/retrieving data by key. Perhaps a wrapper around Api client that utilises session could be good pattern if there are other uses for caching static data?

PR #304, opened 2019-10-15
Apr 722 pya continuity plan

MarkFCain:


src/RoatpWorkflow-QnA.sql
```diff
@@ -5118,17 +5117,42 @@ VALUES
...
+			"Details": {
+        "Title": "What is a significant event?",
+        "Body": "<p class=\"govuk-body\">It''s when something happens that impacts the delivery of apprenticeship training. This includes things like a training location being unavailable or loss of systems and data.</p>"
+      },
 			 "Questions": [
         {
           "QuestionId": "PYA-20",
           "QuestionTag": "ProtectApprentice-Continuity",
-          "Label": "Continuity plan for apprenticeship training",
+          "Label": "Upload your organisation’s continuity plan for apprenticeship training",
```
David-B-Read: This should be double apostrophes ('') as per the body text

src/SFA.DAS.ApplyService.InternalApi/Startup.cs
```diff
@@ -206,9 +204,8 @@ private void ConfigureDependencyInjection(IServiceCollection services)
 
             services.AddTransient<IRoatpApiClient, RoatpApiClient>();
             services.AddTransient<IRoatpTokenService, RoatpTokenService>();
-            services.AddTransient<IKeyProvider,PlaceholderKeyProvider>();
-            services.AddTransient<IStorageService,StorageService>();
-            services.AddTransient<IEncryptionService, EncryptionService>();
+            //services.AddTransient<IKeyProvider,PlaceholderKeyProvider>();
```
David-B-Read: Can these be removed?

src/SFA.DAS.ApplyService.Application/Apply/UploadPageAnswersResult.cs
```diff
@@ -0,0 +1,14 @@
...
+
+namespace SFA.DAS.ApplyService.Application.Apply
+{
+    public class UploadPageAnswersResult
+    {
+        public bool ValidationPassed { get; set; }
+        public List<KeyValuePair<string, string>> ValidationErrors { get; set; }
+        public string NextAction { get; set; }
+        public string NextActionId { get; set; }
+    }
```
davidgouge: You can use `SetPageAnswersResponse` from `SFA.DAS.QnA.Api.Types` instead of creating this one.

src/SFA.DAS.ApplyService.InternalApi/Startup.cs
MarkFCain: sure can, they are all about encryption and storage, which is now subcontracted to QnA

PR #307, opened 2019-10-17
Apr 899 rationalise qna database scripts

MarkFCain:


src/SFA.DAS.ApplyService.sln
```diff
@@ -39,10 +39,6 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "SFA.DAS.ApplyService.Data",
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "SFA.DAS.ApplyService.EmailService", "SFA.DAS.ApplyService.EmailService\SFA.DAS.ApplyService.EmailService.csproj", "{8A3BE608-B568-4D87-9F4B-59559CF573E9}"
 EndProject
-Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "SFA.DAS.ApplyService.Storage", "SFA.DAS.ApplyService.Storage\SFA.DAS.ApplyService.Storage.csproj", "{01E5E3F5-2073-412C-9609-85A00B7FE811}"
-EndProject
```
David-B-Read: Assume that these are no longer needed now that QnA API handling the Azure storage?

MarkFCain: That's the result of some sort of merge issue on my local between submaster and APR-722.  Probably best to park this new script until 722 is in submaster.

PR #308, opened 2019-10-21
Apr 758 confirm trustees

David-B-Read:


src/SFA.DAS.ApplyService.Web/Views/Roatp/WhosInControl/ConfirmTrusteesNoDob.cshtml
```diff
@@ -0,0 +1,58 @@
...
+
+            @if (Model.Trustees.TableData.DataRows.Count == 1)
+            {
+                <h1 class="govuk-heading-xl">Confirm your organisation's trustee</h1>
+            }
+            else
+            {
+                <h1 class="govuk-heading-xl">Confirm your organisation's trustees</h1>
+            }
+            <p class="govuk-body">These are the details we found on the Charity Commission for England and Wales.</p>
```
MarkFCain: I can't see any need for this If statement, the else is the same

PR #316, opened 2019-10-29
Apr 974 verification if file already uploaded

MarkFCain:


src/SFA.DAS.ApplyService.Web/Views/Application/Pages/Index.cshtml
```diff
@@ -216,22 +218,41 @@
...
-
+                            var first = true;
                             @foreach (var link in linksConfiguration.Links)
+                            {
+                                if (first && Model.SectionId != 1)
                                 {
-                                <p class="govuk-body govuk-!-margin-bottom-2">
-                                    <a href="@Url.Action(link.Action, link.Controller, new { Model.ApplicationId, sequenceId = Model.SequenceId, sectionId = Model.SectionId, pageId = Model.PageId})">@link.Caption</a>
-                                </p>
+                                    <p class="govuk-body skipbutton" id="skipbutton">
```
David-B-Read: I think we should raise a tech debt story to restructure this so that the css class is in links configuration, rather than hard-coding the first link if section id != 1

src/SFA.DAS.ApplyService.Web/Views/Application/Pages/Index.cshtml
```diff
@@ -216,22 +218,41 @@
 
                         @if (!Model.HideCTA)
                         {
-                            <button type="submit" class="govuk-button">@Model.CTAButtonText</button>
+                            if (Model.PageOfAnswers != null && Model.PageOfAnswers.Count > 0 && Model.PageOfAnswers.First().Answers.Count(x => x.Value.ToLower().Contains(".pdf"))>0)
```
David-B-Read: As we discussed this is being restructured so it checks for type of file upload

src/SFA.DAS.ApplyService.Web/Views/Application/Pages/Index.cshtml
MarkFCain: I'll do that

src/SFA.DAS.ApplyService.Web/Views/Application/Pages/Index.cshtml
MarkFCain: Will do in the story immediately following this APR-721

PR #333, opened 2019-11-07
Apr 545 add section titles

David-B-Read:


src/SFA.DAS.ApplyService.Web/Controllers/Roatp/RoatpWhosInControlApplicationController.cs
```diff
@@ -773,6 +773,24 @@ public async Task<IActionResult> RemovePscDetails(ConfirmRemovePersonInControlVi
...
+            // for determining that application section completed is derived (APR-1008)
+
+            var applicationSequences = await _qnaApiClient.GetSequences(applicationId);
+            var yourOrganisationSequence =
+                applicationSequences.FirstOrDefault(x => x.SequenceId == RoatpWorkflowSequenceIds.YourOrganisation);
+            var yourOrganisationSections = await _qnaApiClient.GetSections(applicationId, yourOrganisationSequence.Id);
+            var whosInControlSection =
+                yourOrganisationSections.FirstOrDefault(x => x.SectionId == RoatpWorkflowSectionIds.YourOrganisation.WhosInControl);
+
+            await _applicationApiClient.MarkSectionAsCompleted(applicationId, whosInControlSection.Id);
```
MarkFCain: just a note: I'm planning to strip out all the MarkSectionAsCompleted stuff real soon. We might need to take care of how to cycle through this removal if branches are still open.

PR #337, opened 2019-11-11
Apr 1008 fix completed issue

MarkFCain:


src/SFA.DAS.ApplyService.Web/Services/RoatpTaskListWorkflowService.cs
```diff
@@ -90,6 +109,25 @@ private  int SectionHasCompletedQuestions(ApplicationSection section)
             return answeredQuestions;
         }
 
+        private string GetSectionText(int completedCount, ApplicationSection section,  bool sequential)
```
David-B-Read: GetSectionText and SectionText appear to be doing pretty much the same thing?

src/SFA.DAS.ApplyService.Web/Services/RoatpTaskListWorkflowService.cs
```diff
@@ -36,7 +37,25 @@ public  string SectionStatus(IEnumerable<ApplicationSequence> applicationSequenc
...
+            // (MarkSectionAsCompleted, IsSectionCompleted,RemoveSectionCompleted), and the table 'ApplicationWorkflow' may be dropped
+            // I will need to double check there are no other uses for this endpoint before doing that
+            var sectionCompleteBasedOnPagesActiveAndComplete = GetSectionText(questionsCompleted, section,sequential);
+            var sectionCompleteBasedOnDatabaseSettingOfIsComplete = SectionText(questionsCompleted, section.SectionCompleted, sequential);
+
+            var sectionText = sectionCompleteBasedOnPagesActiveAndComplete;
+
+            if (sectionCompleteBasedOnDatabaseSettingOfIsComplete != sectionCompleteBasedOnPagesActiveAndComplete)
+            {
+                if (sequenceId == 1)
```
David-B-Read: If this becomes an enduring solution rather than a temporary one then we should use the RoatpWorkflowSequenceIds.YourOrganisation constant

src/SFA.DAS.ApplyService.Web/Services/RoatpTaskListWorkflowService.cs
MarkFCain: They do, but the 'SectionCompleted' is being got from the database and used in the SectionText method, which is still needed for section 1 until we get it to conform.  The follow up bugfix story will remove one of them

PR #342, opened 2019-11-14
APR-1022 - Added ShowTitleAsCapton and associated switch

davidgouge:


alandburns:
Done a different way in RoATP Apply.

PR #345, opened 2019-11-14
Apr 502 get help with this question

David-B-Read:


src/SFA.DAS.ApplyService.Web/Controllers/Roatp/GetHelpController.cs
```diff
@@ -0,0 +1,126 @@
...
+                if (String.IsNullOrWhiteSpace(ukprn))
+                {
+                    ukprn = "Not available";
+                }
+                getHelpQuery.UKPRN = ukprn;
+                var organisationName = applicationDetails?.UkrlpLookupDetails?.ProviderName;
+                if (String.IsNullOrWhiteSpace(organisationName))
+                {
+                    organisationName = "Not available";
+                }
```
MarkFCain: organisationName = string.IsNullOrWhiteSpace(organisationName) ? "Not available" : organisationName;   

a little less code, but don't care if you leave it

PR #350, opened 2019-11-19
Apr 928 finish application

David-B-Read:


src/SFA.DAS.ApplyService.Web/appsettings.json
```diff
@@ -443,6 +475,15 @@
           "Action": "TaskList"
         }
       ]
+    },
+      "PageId": "10005",
```
MarkFCain: you're missing an opening { here I think....

David-B-Read: Good spot! Fixed

PR #355, opened 2019-11-22
Apr 941 submit application

David-B-Read:


src/SFA.DAS.ApplyService.Web/Controllers/IPageViewModel.cs
```diff
@@ -0,0 +1,14 @@
+﻿using System;
```
martin-hornsby: ViewModels shouldn't be in Controllers

src/SFA.DAS.ApplyService.Web/Controllers/Roatp/GetHelpController.cs
```diff
@@ -0,0 +1,126 @@
...
+            var getHelpQuery = new GetHelpWithQuestion();
+
+            if (applicationId.HasValue && applicationId.Value != Guid.Empty)
+            {
+                var sequences = await _qnaApiClient.GetSequences(applicationId.Value);
+                var currentSequence = sequences.Single(x => x.SequenceId == sequenceId);
+                var sections = await _qnaApiClient.GetSections(applicationId.Value, currentSequence.Id);
+                var currentSection = sections.FirstOrDefault(x => x.SectionId == sectionId);
+
+                var page = await _qnaApiClient.GetPage(applicationId.Value, currentSection.Id, pageId);
```
martin-hornsby: You could cut this down by using _qnaApiClient.GetPageBySectionNo

src/SFA.DAS.ApplyService.Web/Controllers/Roatp/RoatpApplicationController.cs
```diff
@@ -62,7 +63,10 @@ public class RoatpApplicationController : Controller
             ISessionService sessionService, IConfigurationService configService, IUserService userService, IUsersApiClient usersApiClient,
```
martin-hornsby: This controller has lots of dependencies. It's getting out of control.... 😢

src/SFA.DAS.ApplyService.Web/Controllers/Roatp/RoatpApplicationController.cs
```diff
@@ -111,6 +116,8 @@ public async Task<IActionResult> Applications(string applicationType)
             
             switch (application.ApplicationStatus)
             {
+                case ApplicationStatus.Submitted:
```
martin-hornsby: Do you need to detect if an Application is Resubmitted? We have this concept in Assessor

src/SFA.DAS.ApplyService.Web/Controllers/Roatp/RoatpApplicationPreambleController.cs
```diff
@@ -35,7 +35,8 @@ public class RoatpApplicationPreambleController : Controller
         private readonly IUsersApiClient _usersApiClient;
         
         private const string ApplicationDetailsKey = "Roatp_Application_Details";
-        
+        private const string GetHelpSubmittedForPageKey = "Roatp_GetHelpSubmitted_{0}";
```
martin-hornsby: This looks more like a format string than a key

src/SFA.DAS.ApplyService.Web/Controllers/Roatp/RoatpApplicationPreambleController.cs
```diff
@@ -346,6 +358,9 @@ public async Task<IActionResult> SelectApplicationRoute()
             {
                 model.ApplicationRouteId = applicationDetails.ApplicationRoute.Id;
             }
+            var getHelpSessionKey = string.Format(GetHelpSubmittedForPageKey, "ApplicationRoute");
```
martin-hornsby: As mentioned above GetHelpSubmittedForPageKey is a format string and not a key. Please fix

src/SFA.DAS.ApplyService.Web/Services/RoatpTaskListWorkflowService.cs
```diff
@@ -1,17 +1,15 @@
...
-using System.Diagnostics;
 using System.Linq;
 using MoreLinq;
-using NPOI.POIFS.Storage;
 using SFA.DAS.ApplyService.Domain.Entities;
 using SFA.DAS.ApplyService.Web.Configuration;
 
 namespace SFA.DAS.ApplyService.Web.Services
 {
     public class RoatpTaskListWorkflowService: IRoatpTaskListWorkflowService
```
martin-hornsby: The methods defined in this class look like they should be static

src/SFA.DAS.ApplyService.Web/ViewModels/Roatp/ApplicationSummaryViewModel.cs
```diff
@@ -0,0 +1,39 @@
...
+                        }
+                    case "2":
+                        {
+                            return "Employer";
+                        }
+                    case "3":
+                        {
+                            return "Supporting";
+                        }
+                }
```
martin-hornsby: There's no default case

src/SFA.DAS.ApplyService.Web/ViewModels/Roatp/TaskListViewModel.cs
```diff
@@ -157,5 +125,23 @@ public string DescribeOrganisationStartPageId
...
         }
+
+        public bool ApplicationSequencesCompleted()
+        {
+            var applicationSequences = ApplicationSequences.Where(seq => seq.SequenceId != RoatpWorkflowSequenceIds.Finish);
+            foreach (var sequence in applicationSequences)
+            {
+                foreach(var section in sequence.Sections)
+                {
+                    var sectionStatus = SectionStatus(sequence.SequenceId, section.SectionId).ToLower();
```
martin-hornsby: are you able to use section.SequenceId and flatten this foreach loop?

src/SFA.DAS.ApplyService.Web/Controllers/Roatp/RoatpApplicationController.cs
David-B-Read: As discussed, will raise a tech debt story to review this, splitting up and removing redundant code inherited from EPAO

src/SFA.DAS.ApplyService.Web/Controllers/Roatp/RoatpApplicationController.cs
David-B-Read: I don't think we have a concept of Resubmitted in our workflow. They can respond to feedback but any significant changes would be a new application.

src/SFA.DAS.ApplyService.Web/ViewModels/Roatp/TaskListViewModel.cs
David-B-Read: No, the hierarchy is Sequences -> Sections and it needs to iterate through all to ensure that everything mandatory has been answered

src/SFA.DAS.ApplyService.Web/Controllers/Roatp/GetHelpController.cs
David-B-Read: Ok, will raise a tech debt story to implement that in the QnA Api Client and replace occurrences

src/SFA.DAS.ApplyService.Web/Controllers/IPageViewModel.cs
David-B-Read: Inherited tech debt ;) but moved.

src/SFA.DAS.ApplyService.Web/Controllers/Roatp/RoatpApplicationPreambleController.cs
David-B-Read: Renamed constant name

src/SFA.DAS.ApplyService.Web/Controllers/Roatp/RoatpApplicationPreambleController.cs
David-B-Read: Renamed constant name

src/SFA.DAS.ApplyService.Web/Services/RoatpTaskListWorkflowService.cs
David-B-Read: Agreed, converted to static and removed DI

src/SFA.DAS.ApplyService.Web/ViewModels/Roatp/ApplicationSummaryViewModel.cs
David-B-Read: Fixed, in case we add new routes in the future

src/SFA.DAS.ApplyService.Web/ViewModels/Roatp/TaskListViewModel.cs
martin-hornsby: I was thinking more on the lines of: applicationSequences.SelectMany(as => as.Sections)

src/SFA.DAS.ApplyService.Web/ViewModels/Roatp/ApplicationSummaryViewModel.cs
martin-hornsby: it's still dropping out of the case statement and returning String.Empty ...

PR #379, opened 2019-12-10
Apr 1116 tasklist sequence status fixes

David-B-Read:


src/SFA.DAS.ApplyService.Web/ViewModels/Roatp/TaskListViewModel.cs
```diff
@@ -28,6 +27,11 @@ public class TaskListViewModel : ApplicationSummaryViewModel
 
         public IEnumerable<ApplicationSequence> ApplicationSequences { get; set; }
         
+        public TaskListViewModel(IQnaApiClient qnaApiClient)
+        {
+            _qnaApiClient = qnaApiClient;
+        }
+
         public string CssClass(int sequenceId, int sectionId)
```
martin-hornsby: CssClass by normal convention implies Cascading Style Sheets... can we put a comment on this to say it's not that?

src/SFA.DAS.ApplyService.Web/ViewModels/Roatp/TaskListViewModel.cs
```diff
@@ -48,11 +52,183 @@ public string SectionStatus(int sequenceId, int sectionId)
             return RoatpTaskListWorkflowService.SectionStatus(ApplicationSequences, NotRequiredOverrides, sequenceId, sectionId, ApplicationRouteId);
         }
 
+        public string WhosInControlCss
+        {
+            get
+            {
+                var status = WhosInControlSectionStatus;
+
+                if (status == String.Empty)
```
martin-hornsby: what if it's null or whitespace?

src/SFA.DAS.ApplyService.Web/ViewModels/Roatp/TaskListViewModel.cs
```diff
@@ -48,11 +52,183 @@ public string SectionStatus(int sequenceId, int sectionId)
...
+                {
+                    return "hidden";
+                }
+
+                var cssClass = status.ToLower();
```
martin-hornsby: why does it need to be changed to lowercase?

src/SFA.DAS.ApplyService.Web/ViewModels/Roatp/TaskListViewModel.cs
```diff
@@ -48,11 +52,183 @@ public string SectionStatus(int sequenceId, int sectionId)
...
+                cssClass = cssClass.Replace(" ", "");
```
martin-hornsby: do you need to care about tabs? Surely it'll be better to use: cssClass.Trim()

src/SFA.DAS.ApplyService.Web/ViewModels/Roatp/TaskListViewModel.cs
```diff
@@ -48,11 +52,183 @@ public string SectionStatus(int sequenceId, int sectionId)
...
+        public string WhosInControlSectionStatus
+        {
+            get
+            {
+                if (VerifiedCompaniesHouse && VerifiedCharityCommission)
+                {
+                    if ((CompaniesHouseDataConfirmed && !CharityCommissionDataConfirmed)
+                        || (!CompaniesHouseDataConfirmed && CharityCommissionDataConfirmed))
+                    {
+                        return "In Progress";
```
martin-hornsby: Hard coded magic strings. Please put these as constants or an Enum Class somewhere

src/SFA.DAS.ApplyService.Web/ViewModels/Roatp/TaskListViewModel.cs
```diff
@@ -48,11 +52,183 @@ public string SectionStatus(int sequenceId, int sectionId)
...
+                    }
+                    if (CompaniesHouseDataConfirmed && CharityCommissionDataConfirmed)
```
martin-hornsby: I'm not sure all of these if statements are efficient. Couldn't you use if/else if or move to a switch ?

Are there unit tests to cover all of this logic ?

src/SFA.DAS.ApplyService.Web/ViewModels/Roatp/TaskListViewModel.cs
```diff
@@ -48,11 +52,183 @@ public string SectionStatus(int sequenceId, int sectionId)
...
+        {
+            var status = FinishSectionStatus(sectionId);
+
+            if (status == String.Empty)
+            {
+                return "hidden";
+            }
+
+            var cssClass = status.ToLower();
+            cssClass = cssClass.Replace(" ", "");
```
martin-hornsby: See comments above

src/SFA.DAS.ApplyService.Web/ViewModels/Roatp/TaskListViewModel.cs
David-B-Read: Re-written to use TaskListSectionStatus constants

src/SFA.DAS.ApplyService.Web/ViewModels/Roatp/TaskListViewModel.cs
David-B-Read: This is because the CSS class names are lowercase, per convention, and are case-sensitive

src/SFA.DAS.ApplyService.Web/ViewModels/Roatp/TaskListViewModel.cs
David-B-Read: This code was used to map statuses like In Progress and Not Required to their relevant CSS class names. I've rewritten this code so that it's easier to see what the mappings are

src/SFA.DAS.ApplyService.Web/ViewModels/Roatp/TaskListViewModel.cs
David-B-Read: Done

src/SFA.DAS.ApplyService.Web/ViewModels/Roatp/TaskListViewModel.cs
David-B-Read: There are unit tests for this. The code has to cover cases where a provider can be a company, a charity or and company and a charity.

src/SFA.DAS.ApplyService.Web/ViewModels/Roatp/TaskListViewModel.cs
David-B-Read: CSS mapping code has been rewritten so it's clearer

src/SFA.DAS.ApplyService.Web/ViewModels/Roatp/TaskListViewModel.cs
```diff
@@ -64,15 +215,15 @@ public bool IntroductionPageNextSectionUnavailable(int sequenceId, int sectionId
                 foreach(var section in yourOrganisationSequence.Sections)
                 {
                     var sectionStatus = RoatpTaskListWorkflowService.SectionStatus(ApplicationSequences, NotRequiredOverrides, RoatpWorkflowSequenceIds.YourOrganisation, section.SectionId, ApplicationRouteId);
-                    if (sectionStatus.ToLower() != PageStatusCompleted)
+                    if (sectionStatus != TaskListSectionStatus.Completed)
```
martin-hornsby: Sorry, just seen this. To make it safer I would do  if (!TaskListSectionStatus.Completed.Equals(sectionStatus, StringComparison.InvariantCultureIgnoreCase))

src/SFA.DAS.ApplyService.Web/ViewModels/Roatp/TaskListViewModel.cs
```diff
@@ -64,15 +215,15 @@ public bool IntroductionPageNextSectionUnavailable(int sequenceId, int sectionId
...
                     {
                         return true;
                     }
                 }
             }
 
             var statusOfIntroductionPage = SectionStatus(sequenceId,IntroductionSectionId);
-            if (sequenceId > Sequence1Id && sectionId != IntroductionSectionId && statusOfIntroductionPage.ToLower() != PageStatusCompleted)
+            if (sequenceId > Sequence1Id && sectionId != IntroductionSectionId && statusOfIntroductionPage != TaskListSectionStatus.Completed)
```
martin-hornsby: See above comment

src/SFA.DAS.ApplyService.Web/ViewModels/Roatp/TaskListViewModel.cs
```diff
@@ -133,8 +288,8 @@ public bool ApplicationSequencesCompleted()
             {
                 foreach(var section in sequence.Sections)
                 {
-                    var sectionStatus = SectionStatus(sequence.SequenceId, section.SectionId).ToLower();
-                    if (sectionStatus != "not required" && sectionStatus != "completed")
+                    var sectionStatus = SectionStatus(sequence.SequenceId, section.SectionId);
+                    if (sectionStatus != TaskListSectionStatus.NotRequired && sectionStatus != TaskListSectionStatus.Completed)
```
martin-hornsby: See above comment

src/SFA.DAS.ApplyService.Web/ViewModels/Roatp/TaskListViewModel.cs
```diff
@@ -143,5 +298,23 @@ public bool ApplicationSequencesCompleted()
 
             return true;
         }
+
+        private string ConvertTaskListSectionStatusToCssClass(string sectionStatus)
+        {
+            if (sectionStatus == TaskListSectionStatus.Blank)
```
martin-hornsby: this should be a switch statement

David-B-Read: Case only works with numerics, so converted to an else if

src/SFA.DAS.ApplyService.Web/ViewModels/Roatp/TaskListViewModel.cs
David-B-Read: The SectionStatus function now only returns one of the TaskListSectionStatus constant values, so case checking not required here

PR #402, opened 2020-01-10
Apr 1014 change ukprn

David-B-Read:


src/SFA.DAS.ApplyService.Application/Apply/Roatp/UpdateApplicationStatusHandler.cs
```diff
@@ -0,0 +1,27 @@
...
+        {
+            _logger = logger;
+            _repository = repository;
+        }
+
+        protected override bool Handle(UpdateApplicationStatusRequest request)
+        {
+            _logger.LogInformation($"Updating application status to {request.ApplicationStatus} for application ID {request.ApplicationId}");
+
+            _repository.UpdateApplicationStatus(request.ApplicationId, request.ApplicationStatus);
```
MarkFCain: this is optional.... maybe wrap in a try catch error and only proceed to 'true' if it's all ok?

David-B-Read: I think it should log the failure and proceed, the key outcome is that the user is able to restart their application from scratch. Any issues with updating could be picked up as a data fix

PR #405, opened 2020-01-13
Apr 1015 change route interstitial page

David-B-Read:


src/SFA.DAS.ApplyService.Web/Controllers/Roatp/RoatpApplicationPreambleController.cs
```diff
@@ -335,7 +335,7 @@ public async Task<IActionResult> SelectApplicationRoute()
             model.ApplicationRoutes = applicationRoutes;
 
             var applicationDetails = _sessionService.Get<ApplicationDetails>(ApplicationDetailsKey);
-            if (applicationDetails.ApplicationRoute != null)
+            if (applicationDetails != null && applicationDetails.ApplicationRoute != null)
```
MarkFCain: simplify to applicationDetails?.ApplicationRoute !=null

PR #412, opened 2020-01-22
APR-1152 store section exemption rules to ApplyData JSON

David-B-Read:


alandburns:
This was a breaking change, so needs to be resolved before re-introducing.

PR #418, opened 2020-02-03
APR-1268 - Dummy RoATP Gateway Review

martin-hornsby:


src/SFA.DAS.ApplyService.Domain/Apply/GatewayApplicationSummaryItem.cs
```diff
@@ -0,0 +1,17 @@
+﻿using System;
+
+namespace SFA.DAS.ApplyService.Domain.Apply
+{
+    public class GatewayApplicationSummaryItem
```
David-B-Read: This looks to be the same information displayed across Gateway, FHA and Assessor views, so wondered if this could become more generic?

martin-hornsby: Make this generic so we now have RoatpApplicationSummaryItem

PR #421, opened 2020-02-05
remove unused tables

alandburns:

PR #424, opened 2020-02-06
Apr 1269 fha review

David-B-Read:


src/SFA.DAS.ApplyService.Application/Apply/Financial/RecordGradeRequest.cs
```diff
@@ -0,0 +1,19 @@
...
+using SFA.DAS.ApplyService.Domain.Entities;
+using System;
+
+namespace SFA.DAS.ApplyService.Application.Apply.Financial
+{
+    public class RecordGradeRequest : IRequest<bool>
+    {
+        public Guid ApplicationId { get; set; }
+        public FinancialReviewDetails FinancialReviewDetails { get; set; }
+        public string FinancialReviewStatus { get; set; }
```
martin-hornsby: As mentioned in the other PR. Would expect this to be calculated within the Handler

src/SFA.DAS.ApplyService.Application/Apply/Financial/StartFinancialReviewHandler.cs
```diff
@@ -1,29 +1,27 @@
...
-            var section = await _applyRepository.GetSection(request.ApplicationId, 1, 3, null);
-
-            if (section.Status == ApplicationSectionStatus.Submitted)
-            {
-                await _applyRepository.StartFinancialReview(request.ApplicationId);   
-            }
-            
-            return Unit.Value;
+            _logger.LogInformation($"Starting financial review for application {request.ApplicationId}");
+            await _applyRepository.StartFinancialReview(request.ApplicationId, request.Reviewer);
```
martin-hornsby: You've removed the guard check... it was put in place to stop an application getting into a bad state due to an incorrect API call

src/SFA.DAS.ApplyService.Application/Apply/IApplyRepository.cs
```diff
@@ -33,12 +41,12 @@ public interface IApplyRepository
...
-        Task<List<Domain.Entities.Apply>> GetClosedApplications();
-        Task<List<Domain.Entities.Apply>> GetOpenFinancialApplications();
-        Task<List<Domain.Entities.Apply>> GetFeedbackAddedFinancialApplications();
-        Task<List<Domain.Entities.Apply>> GetClosedFinancialApplications();        
+        Task<List<RoatpApplicationSummaryItem>> GetOpenApplications();
+        Task<List<RoatpApplicationSummaryItem>> GetFeedbackAddedApplications();
+        Task<List<RoatpApplicationSummaryItem>> GetClosedApplications();
+        Task<List<RoatpFinancialSummaryItem>> GetOpenFinancialApplications();
+        Task<List<RoatpFinancialSummaryItem>> GetFeedbackAddedFinancialApplications();
+        Task<List<RoatpFinancialSummaryItem>> GetClosedFinancialApplications();
```
martin-hornsby: Can these be moved above the comment saying '// NOTE: This is old stuff or things which are not migrated over yet'

src/SFA.DAS.ApplyService.Application/Apply/IApplyRepository.cs
```diff
@@ -63,5 +70,11 @@ public interface IApplyRepository
         Task<IEnumerable<RoatpApplicationStatus>> GetExistingApplicationStatusByUkprn(string ukprn);
 
         Task<string> GetNextRoatpApplicationReference();
+
+        Task<bool> RecordFinancialGrade(Guid applicationId, FinancialReviewDetails financialReviewDetails, string financialReviewStatus);
+
+        Task<bool> StartAssessorReview(Guid applicationId, string reviewer);
+
+        Task<bool> StartFinancialReview(Guid applicationId, string reviewer);
```
martin-hornsby: Can these be moved above the comment saying '// NOTE: This is old stuff or things which are not migrated over yet'

src/SFA.DAS.ApplyService.Database/Script.PostDeployment1.sql
```diff
@@ -11,95 +11,7 @@ Post-Deployment Script Template
...
-
-IF NOT EXISTS (SELECT * FROM EmailTemplates WHERE TemplateName = N'ApplyEPAOResponse')
-BEGIN
-INSERT EmailTemplates ([Id], [Status],[TemplateName],[TemplateId],[CreatedAt],[CreatedBy]) 
-VALUES (NEWID(), 'Live', N'ApplyEPAOResponse', N'84174eab-f3c1-4274-8670-2fb5b21cbd77', GETDATE(), 'System')  
-END
-
-DELETE FROM EmailTemplates
-WHERE [TemplateName] IN ( 'EPAOUserApproveRequest' , 'EPAOUserApproveConfirm' )
+DELETE FROM [EmailTemplates]
```
martin-hornsby: Are we sure we should clearing out all of the EmailTemplates?

src/SFA.DAS.ApplyService.Database/Script.PreDeployment1.sql
```diff
@@ -31,4 +31,36 @@ IF OBJECT_ID('dbo.QnAs', 'U') IS NOT NULL
...
+IF (EXISTS (SELECT * FROM INFORMATION_SCHEMA.TABLES WHERE TABLE_SCHEMA = 'dbo' AND  TABLE_NAME = 'WorkflowSections'))
+    DROP TABLE WorkflowSections;
+IF (EXISTS (SELECT * FROM INFORMATION_SCHEMA.TABLES WHERE TABLE_SCHEMA = 'dbo' AND  TABLE_NAME = 'Workflows'))
+    DROP TABLE Workflows;
+IF (EXISTS (SELECT * FROM INFORMATION_SCHEMA.TABLES WHERE TABLE_SCHEMA = 'dbo' AND  TABLE_NAME = 'Assets'))
+    DROP TABLE Assets;
+IF (EXISTS (SELECT * FROM INFORMATION_SCHEMA.TABLES WHERE TABLE_SCHEMA = 'dbo' AND  TABLE_NAME = 'QnAs'))
+    DROP TABLE QnAs;
+IF (EXISTS (SELECT * FROM INFORMATION_SCHEMA.TABLES WHERE TABLE_SCHEMA = 'dbo' AND  TABLE_NAME = 'Entities'))
+    DROP TABLE Entities;
```
martin-hornsby: Doesn't the SQL Dacpac remove these tables if they are not within the publish script?

src/SFA.DAS.ApplyService.InternalApi/Controllers/FinancialController.cs
```diff
@@ -48,9 +56,9 @@ public async Task<ActionResult> UpdateGrade(Guid applicationId, [FromBody] Finan
         }
 
         [HttpPost("/Financial/{applicationId}/StartReview")]
-        public async Task<ActionResult> StartReview(Guid applicationId)
+        public async Task<ActionResult> StartReview(Guid applicationId, [FromBody] string reviewer)
```
martin-hornsby: Does this work? Normally would have expected a Request object. What if you wanted to expand on it later down the line? Have a look at what I did for Gateway as this was the approach taken within EPAO

src/SFA.DAS.ApplyService.Application/Apply/Financial/RecordGradeRequest.cs
David-B-Read: Done

src/SFA.DAS.ApplyService.Database/Script.PostDeployment1.sql
David-B-Read: This is from PR
https://github.com/SkillsFundingAgency/das-apply-service/pull/421

src/SFA.DAS.ApplyService.Database/Script.PreDeployment1.sql
David-B-Read: This is from PR
https://github.com/SkillsFundingAgency/das-apply-service/pull/421

src/SFA.DAS.ApplyService.Application/Apply/Financial/StartFinancialReviewHandler.cs
David-B-Read: The SQL query prevents the state from being changed to an invalid value

src/SFA.DAS.ApplyService.Application/Apply/IApplyRepository.cs
David-B-Read: Done

src/SFA.DAS.ApplyService.Application/Apply/IApplyRepository.cs
David-B-Read: Done

src/SFA.DAS.ApplyService.InternalApi/Controllers/FinancialController.cs
David-B-Read: Updated this, and the assessor review, to use Request objects so consistent with Gateway checks

src/SFA.DAS.ApplyService.Application/Apply/Financial/StartFinancialReviewHandler.cs
David-B-Read: At present, there aren't section/sequence level statuses being set. The old code was retrieving the EPAO sequence/section from the database so was invalid for RoATP.

martin-hornsby: Fair enough. Marking as resolved

src/SFA.DAS.ApplyService.InternalApi/Controllers/FinancialController.cs
```diff
@@ -40,6 +41,13 @@ public async Task<ActionResult> ClosedApplications()
             return Ok(applications);
         }
 
+        [HttpPost("/Financial/Grade/{applicationId}")]
+        public async Task<IActionResult> RecordGrade(Guid applicationId, [FromBody] FinancialReviewDetails financialReviewDetails)
+        {
+            await _mediator.Send(new RecordGradeRequest(applicationId, financialReviewDetails));
+            return Ok();
+        }
+
```
martin-hornsby: ahhh just noticed this. If I remember correct this is what UpdateGrade would have done.

As a minimum can you change the Route to: [HttpPost("/Financial/{applicationId}/Grade")]
This will make it consistent with all the other end points

src/SFA.DAS.ApplyService.Data/ApplyRepository.cs
```diff
@@ -121,16 +120,156 @@ public async Task SubmitApplication(Guid applicationId, ApplyData applyData, Gui
             using (var connection = new SqlConnection(_config.SqlConnectionString))
             {
                 await connection.ExecuteAsync(@"UPDATE Apply
-                                                SET  ApplicationStatus = @ApplicationStatus, ApplyData = @applyData, ReviewStatus = @ReviewStatus, UpdatedBy = @submittedBy, UpdatedAt = GETUTCDATE() 
+                                                SET  ApplicationStatus = @ApplicationStatus, ApplyData = @applyData, AssessorReviewStatus = @ReviewStatus, GatewayReviewStatus = @GatewayReviewStatus, UpdatedBy = @submittedBy, UpdatedAt = GETUTCDATE()
```
martin-hornsby: Looks like FinancialReviewStatus is missing

src/SFA.DAS.ApplyService.Data/ApplyRepository.cs
```diff
@@ -796,5 +980,95 @@ public async Task<string> GetNextRoatpApplicationReference()
...
             }
         }
+
+        public async Task<bool> RecordFinancialGrade(Guid applicationId, FinancialReviewDetails financialReviewDetails, string financialReviewStatus)
+        {
+            using (var connection = new SqlConnection(_config.SqlConnectionString))
+            {
+                try
+                {
+                    var result = await connection.QueryAsync<ApplyData>(@"SELECT ApplyData FROM Apply WHERE ApplicationId = @applicationId",
```
martin-hornsby: More of a comment but I'd have personally used GetApplication() than have a separate Dapper query. That's just so retrieving application info is done in one place, but then you can argue that your way brings back a lighter object

src/SFA.DAS.ApplyService.Data/ApplyRepository.cs
```diff
@@ -796,5 +980,95 @@ public async Task<string> GetNextRoatpApplicationReference()
...
+                                                                        WHERE ApplicationId = @applicationId",
+                        new
+                        {
+                            applicationId,
+                            applyData,
+                            grade = financialReviewDetails.SelectedGrade
+                        });
+                    if (recordsAffected > 0)
+                    {
+                        await connection.ExecuteAsync("UPDATE Apply SET FinancialReviewStatus = @financialReviewStatus WHERE ApplicationId = @applicationId",
```
martin-hornsby: Does this need to be done as a separate call, especially since you know that ApplyData already exists and thus will be recordsAffected will equal 1? If so can we have a comment in the code

src/SFA.DAS.ApplyService.Data/ApplyRepository.cs
```diff
@@ -796,5 +980,95 @@ public async Task<string> GetNextRoatpApplicationReference()
...
+            using (var connection = new SqlConnection(_config.SqlConnectionString))
+            {
+                await connection.ExecuteAsync(@"UPDATE Apply SET FinancialReviewStatus = @financialReviewStatus,
+                                                UpdatedAt = @updatedAt, UpdatedBy = @updatedBy
+                                                WHERE ApplicationId = @applicationId
+                                                AND FinancialReviewStatus IN ( @draftStatus, @newStatus )",
+                        new
+                        {
+                            applicationId,
+                            financialReviewStatus = FinancialReviewStatus.InProgress,
```
martin-hornsby: could this say financialReviewStatusInProgress so it's slightly more clear on what the status is being set to?

src/SFA.DAS.ApplyService.Data/ApplyRepository.cs
David-B-Read: It would be set to the default value defined in the table definition, but it's a good idea to document what the initial state of the application is in code.

src/SFA.DAS.ApplyService.Data/ApplyRepository.cs
David-B-Read: Good call, compressed into a single update query

src/SFA.DAS.ApplyService.Data/ApplyRepository.cs
David-B-Read: Done, and the start assessor review status variable too

src/SFA.DAS.ApplyService.Data/ApplyRepository.cs
David-B-Read: Yeah, I did try and do the update in a single step here using JSON_MODIFY but found that it set the field value to serialised JSON rather than an object, it seems that only works for simple types.

src/SFA.DAS.ApplyService.InternalApi/Controllers/FinancialController.cs
David-B-Read: Done

PR #427, opened 2020-02-07
Section76 sectors and employee experience

MarkFCain:


src/SFA.DAS.ApplyService.Web/Controllers/Roatp/RoatpApplicationController.cs
```diff
@@ -844,7 +852,7 @@ private static void ApplyFormattingToAnswers(List<Answer> answers, Page page)
 
         private static IEnumerable<Question> PageContainsCheckBoxListQuestions(Page page)
         {
-            return page.Questions.Where(q => q.Input.Type == "CheckBoxList");
+            return page.Questions.Where(q => q.Input.Type == "CheckBoxList" || q.Input.Type=="ComplexCheckBoxList");
```
David-B-Read: Could this be simplified to .Contains("CheckBoxList") ?

src/SFA.DAS.ApplyService.Web/Controllers/Roatp/RoatpApplicationController.cs
```diff
@@ -890,18 +899,47 @@ private List<Answer> GetAnswersFromForm(Page page)
...
+                        foreach (var answer in answers.Where(a => a.QuestionId == furtherQuestion.QuestionId))
+                        {
+                            answer.Value = string.Empty;
+                        }
+                    }
+                }
 
-                // Remove FurtherQuestion answers to all other Options as they were not selected and thus should not be stored
-                foreach (var furtherQuestion in question.Input.Options.Where(opt => opt.Value != answerForQuestion?.Value && opt.FurtherQuestions != null).SelectMany(opt => opt.FurtherQuestions))
+                if (question.Input.Type == "ComplexCheckBoxList")
```
David-B-Read: Maybe re-write as an else if or case statement, as a question will only have one input type?

src/SFA.DAS.ApplyService.Web/Infrastructure/Services/PagesWithSectionsFlowService.cs
```diff
@@ -0,0 +1,43 @@
...
+using System.Linq;
+using System.Threading.Tasks;
+using SFA.DAS.ApplyService.Domain.Apply;
+using SFA.DAS.ApplyService.Domain.Entities;
+using SFA.DAS.ApplyService.Web.Infrastructure.Interfaces;
+using SFA.DAS.ApplyService.Web.ViewModels.Roatp;
+
+namespace SFA.DAS.ApplyService.Web.Infrastructure.Services
+{
+    public class PagesWithSectionsFlowService: IPagesWithSectionsFlowService
```
David-B-Read: As discussed, this is a good candidate for some unit test coverage in the tech debt story

src/SFA.DAS.ApplyService.Web/Views/Application/PagesWithSections.cshtml
```diff
@@ -1,15 +1,20 @@
...
+@model SFA.DAS.ApplyService.Domain.Entities.ApplicationSection
 
 @{
     ViewBag.Title = @Model.Title;
     Layout = "_Layout";
     var pageTitle = "";
     var isSamePage = false;
     var nextPageTitle = "";
-    var pagesToCheck = Model.QnAData.Pages.Where(p => p.Active && !p.NotRequired).ToList();
+    var pagesToCheck = Model.QnAData.Pages.Where(p => p.Active && !p.NotRequired && p.DisplayType!=null && p.DisplayType == "PagesWithSections").ToList();
```
David-B-Read: Possibly write this as a method on the view model, rather than logic in the view

src/SFA.DAS.ApplyService.Web/Controllers/Roatp/RoatpApplicationController.cs
```diff
@@ -8,6 +8,7 @@
 using Microsoft.Extensions.Logging;
 using Newtonsoft.Json;
 using Newtonsoft.Json.Linq;
+using NPOI.SS.Formula.Functions;
```
martin-hornsby: What's this NuGet package for?

src/SFA.DAS.ApplyService.Web/Controllers/Roatp/RoatpApplicationController.cs
martin-hornsby: Please make it explicit

src/SFA.DAS.ApplyService.Web/Controllers/Roatp/RoatpApplicationController.cs
martin-hornsby: This could be written into a more simple function as it's only this bit that is different. Would save a lot of copy & paste noise:

`foreach (var answer in answers.Where(a => a.QuestionId == furtherQuestion.QuestionId))
                        {
                            answer.Value = string.Empty;
                        }`

src/SFA.DAS.ApplyService.Web/Views/Application/PagesWithSections.cshtml
martin-hornsby: A lambda property on the viewmodel would work :)

src/SFA.DAS.ApplyService.Web/Controllers/Roatp/RoatpApplicationController.cs
martin-hornsby: Nope. It should stay as being an explicit list. New input types will appear over time and doing a contains may introduce dodgy behaviour.

Imagine for example 'ComplexRadioWithCheckBoxList"

src/SFA.DAS.ApplyService.Web/Controllers/Roatp/RoatpApplicationController.cs
martin-hornsby: On further inspection it's not quite the same logic.

PR #428, opened 2020-02-07
APR-1194 Update submit application page - feedback from legal

MakCom:


src/SFA.DAS.ApplyService.Web/ViewModels/Roatp/SubmitApplicationViewModel.cs
```diff
@@ -9,8 +9,10 @@ public class SubmitApplicationViewModel
     {
         public Guid ApplicationId { get; set; }
         public string OrganisationName { get; set; }
-        [Required(ErrorMessage = "You must confirm that all the answers and uploads in this application are accurate and to the best of your knowledge")]
-        public string ConfirmSubmitApplication { get; set; }
+        [Range(typeof(bool), "true", "true", ErrorMessage = "Tell us if you confirm that all the answers and uploads in this application are accurate and to the best of your knowledge")]
```
martin-hornsby: nice hack on the data annotation 😉  it works without the need for a custom one 👍

David-B-Read: Is the copy in this error message correct? Looks like an extra “and” at the end

MakCom: I have discussed the issue of the "extra end" with Priya and the text is updated now.

PR #429, opened 2020-02-10
APR-1278 - RoATP Apply Snapshot

martin-hornsby:


src/SFA.DAS.ApplyService.Data/ApplyRepository.cs
```diff
@@ -135,6 +135,42 @@ public async Task SubmitApplication(Guid applicationId, ApplyData applyData, Gui
...
+            {
+                var newApplyData = new ApplyData
+                {
+                    Sequences = newSequences,
+                    ApplyDetails = currentApplication.ApplyData?.ApplyDetails
+                };
+
+                using (var connection = new SqlConnection(_config.SqlConnectionString))
+                {
+                    // todo: CHECK updated SQL
```
MarkFCain: what's this todo all about?

martin-hornsby: good spot - removed it

PR #440, opened 2020-02-11
APR-1193 Update Finish to remove CoAs and add IfATE and on-boarding questions

MakCom:


src/SFA.DAS.ApplyService.Web/ViewModels/Roatp/TaskListViewModel.cs
```diff
@@ -238,7 +250,16 @@ public bool PreviousSectionCompleted(int sequenceId, int sectionId)
                 {
                     return true;
                 }
-                return (FinishSectionStatus(sectionId-1) == TaskListSectionStatus.Completed);
+
+                // TODO: APR-1193 - I have added the check for the case of section 9.3 being NotRequired for Supporting Provider
```
David-B-Read: This is a brittle solution that would need code changes in Apply if the sections were re-organised in QnA. It would be simpler to check for previous sections having a status of Completed or Not Required

src/SFA.DAS.ApplyService.Web/ViewModels/Roatp/TaskListViewModel.cs
```diff
@@ -167,8 +177,10 @@ public string FinishSectionStatus(int sectionId)
                 }
                 else if (ApplicationRouteId == SupportingApplicationRouteId)
                 {
-                    conditionsOfAcceptance2 = _qnaApiClient.GetAnswerByTag(ApplicationId, RoatpWorkflowQuestionTags.FinishCOA2Supporting).GetAwaiter().GetResult();
-                    conditionsOfAcceptance3 = _qnaApiClient.GetAnswerByTag(ApplicationId, RoatpWorkflowQuestionTags.FinishCOA3Supporting).GetAwaiter().GetResult();
+                    // TODO: TODO: APR-1193 - This is the return value for 9.3 section for Supporting Provider
```
David-B-Read: There's a NotRequiredOverrides section in config where rules are configured based on previously entered answers, can this logic be moved to there?

MakCom: Yes, there was a code in the NotRequiredOverrides section in appsettings.json, which got overridden by the merge.
I have restored it.
But I intent to keep the code here as it is really reflecting the conditions of all states  

  {
      "SequenceId": "9",
      "SectionId": "3",
      "Conditions": [
        {
          "ConditionalCheckField": "Apply-ProviderRoute",
          "MustEqual": "3"
        }
      ]
    }

src/SFA.DAS.ApplyService.Web/ViewModels/Roatp/TaskListViewModel.cs
MakCom: Checking the status Completed or Not Required in this case is not enough. We have to take into consideration Supporting Provider check. And the same approach is used for all other case. 
Any refactoring should be done for the whole methods, which will involves significant amount of coding and regressive testing.

alandburns: See https://skillsfundingagency.atlassian.net/browse/APR-1356 
This needs to be properley resolved ASAP.
For now I will approve fior testing.

src/SFA.DAS.ApplyService.Web/ViewModels/Roatp/TaskListViewModel.cs
alandburns: See https://skillsfundingagency.atlassian.net/browse/APR-1356 
This needs to be properley resolved ASAP.
For now I will approve fior testing.

PR #443, opened 2020-02-12
Do not want to delete the contents of the Apply Table.

alandburns:


src/SFA.DAS.ApplyService.Database/Script.PreDeployment1.sql
```diff
@@ -33,16 +33,6 @@ IF OBJECT_ID('dbo.Entities', 'U') IS NOT NULL
...
-IF NOT EXISTS (
-  SELECT
-    *
-  FROM
-    INFORMATION_SCHEMA.COLUMNS
-  WHERE
-    TABLE_NAME = 'Apply' AND COLUMN_NAME = 'AssessorReviewStatus')
-BEGIN
-    DELETE FROM dbo.Apply
-END
```
David-B-Read: Will require DACPAC deploy settings to be changed by DevOps, as schema changes are prevented when tables are populated by default

alandburns: Yes. But this is very dangerous. And it is safer to have DevOps override DACPAC warnings.
And if only adding a column it should be fine anyway.

PR #448, opened 2020-02-13
APR-1355 : Can Update Application check

martin-hornsby:


src/SFA.DAS.ApplyService.Data/ApplyRepository.cs
```diff
@@ -40,10 +40,10 @@ public async Task<Guid> StartApplication(Guid applicationId, ApplyData applyData
             using (var connection = new SqlConnection(_config.SqlConnectionString))
             {
                 return await connection.QuerySingleAsync<Guid>(
-                    @"INSERT INTO Apply (ApplicationId, OrganisationId, ApplicationStatus, ApplyData, AssessorReviewStatus, CreatedBy, CreatedAt)
+                    @"INSERT INTO Apply (ApplicationId, OrganisationId, ApplicationStatus, ApplyData, CreatedBy, CreatedAt)
```
David-B-Read: Initial assessor review status should be set here

src/SFA.DAS.ApplyService.Web/Controllers/Roatp/RoatpApplicationController.cs
```diff
@@ -503,6 +486,12 @@ public async Task<IActionResult> SaveAnswers(PageViewModel vm, Guid applicationI
         [HttpGet]
         public async Task<IActionResult> TaskList(Guid applicationId)
         {
+            var canUpdate = await CanUpdateApplication(applicationId);
+            if (!canUpdate)
+            {
+                return RedirectToAction("Applications");
```
David-B-Read: Does this view need updating with RoATP specific content? I don't think it's been touched since EPAO

src/SFA.DAS.ApplyService.Web/Controllers/Roatp/RoatpApplicationController.cs
```diff
@@ -1051,9 +1068,9 @@ private bool FileValidationPassed(List<Answer> answers, Page page, List<Validati
         }
 
         [HttpPost]
-        public async Task<IActionResult> Submit(Guid applicationId, int sequenceId)
+        public async Task<IActionResult> Submit(Guid applicationId)
```
David-B-Read: This controller action has been removed in future branches, the check should be done in SubmitApplication / ConfirmSubmitApplication if not done already

src/SFA.DAS.ApplyService.Data/ApplyRepository.cs
martin-hornsby: We'll need to do it with all other Status columns. I removed them as the database will put default ones in

src/SFA.DAS.ApplyService.Web/Controllers/Roatp/RoatpApplicationController.cs
martin-hornsby: Nope, it's the TaskList for RoATP

src/SFA.DAS.ApplyService.Web/Controllers/Roatp/RoatpApplicationController.cs
martin-hornsby: this will have to be done retrospectively

src/SFA.DAS.ApplyService.Data/ApplyRepository.cs
martin-hornsby: Just looked at this... it's using ApplicationStatus from the wrong Entity. I'm refusing to add the other statuses until this is properly sorted out

src/SFA.DAS.ApplyService.Web/Controllers/Roatp/RoatpApplicationController.cs
martin-hornsby: included check in ConfirmSubmitApplication

src/SFA.DAS.ApplyService.Web/Controllers/Roatp/RoatpApplicationController.cs
martin-hornsby: Applications() is the method that is executed once you've logged into RoATP

PR #456, opened 2020-02-17
APR-1361 - Update ApplyDetails.ProviderRoute when changing Provider Route

martin-hornsby:


src/SFA.DAS.ApplyService.Data/ApplyRepository.cs
```diff
@@ -127,7 +127,28 @@ public async Task SubmitApplication(Guid applicationId, ApplyData applyData, Gui
...
+            var application = await GetApplication(applicationId);
+            var applyData = application?.ApplyData;
+
+            if (application != null && applyData?.ApplyDetails != null)
+            {
+                applyData.ApplyDetails.ProviderRoute = ProviderRoute;
 
+                using (var connection = new SqlConnection(_config.SqlConnectionString))
+                {
+                    await connection.ExecuteAsync(@"UPDATE Apply
```
David-B-Read: Can this be simplied to use JSON_MODIFY?

martin-hornsby: yeah it can... done 👍

PR #464, opened 2020-02-21
APR-1428 code done

MarkFCain:


src/SFA.DAS.ApplyService.Web/Controllers/Roatp/RoatpManagementHierarchyController.cs
```diff
@@ -107,7 +107,7 @@ public async Task<IActionResult> AddManagementHierarchyDetails(AddEditManagement
             {
                 managementHierarchyData = new TabularData
                 {
-                    HeadingTitles = new List<string> { "Name", "Job role"},
+                    HeadingTitles = new List<string> { "Name", "Job role","Years in role","Months in role","Part of another organisation","organisation details"},
```
David-B-Read: Capitalisation on the 'organisation details' header?

PR #468, opened 2020-02-25
APR-1396 Stop Inactive Pages being accessed

MakCom:


src/SFA.DAS.ApplyService.Web/Controllers/Roatp/RoatpApplicationController.cs
```diff
@@ -621,7 +621,7 @@ private void PopulateAdditionalSequenceFields(IEnumerable<ApplicationSequence> s
             }
         }
 
-        private async Task<bool> CanUpdateApplication(Guid applicationId, int? sequenceId = null, int? sectionId = null)
+        private async Task<bool> CanUpdateApplication(Guid applicationId, int? sequenceId = null, int? sectionId = null, string pageId = null)
```
David-B-Read: This logic has become quite complex now, so would benefit from being split out into a separate service and covered by unit tests

MichaelYoung1981: +1

PR #471, opened 2020-02-27
APR-1389 Use QnA Api GetSectionBySectionNo end point

MakCom:
This PR/branch contains changes for the following 3 JIRA tickets:
- APR-1389 Use QnA Api GetSectionBySectionNo end point
- APR-1390 Use QnA Api GetSequenceBySequenceNo end point
- APR-1406 Use QnA Api GetPageBySectionNo end point


martin-hornsby:
Just need to check that no regression has occurred

PR #476, opened 2020-03-09
Apr 1283 gateway answer table

MarkFCain:


src/SFA.DAS.ApplyService.Database/Tables/GatewayAnswer.sql
```diff
@@ -0,0 +1,23 @@
+﻿CREATE TABLE [dbo].[GatewayAnswer](
+	[Id] [uniqueidentifier] NOT NULL,
+	[ApplicationId] [uniqueidentifier] NOT NULL,
```
David-B-Read: Might be a good idea to index Application Id and Page Id

src/SFA.DAS.ApplyService.Domain/Apply/Answer.cs
```diff
@@ -16,7 +16,7 @@ public string Value
             set { JsonValue = value; }
         }
 
-        [JsonProperty(PropertyName = "Value")]
+        [JsonProperty(PropertyName = "Status")]
```
David-B-Read: I think this domain object is used in apply, so this might break serialisation elsewhere??

MakCom: It could be directly SectionReviewStatus

src/SFA.DAS.ApplyService.Database/Tables/GatewayAnswer.sql
MarkFCain: Excellent point.  We also use it coupled tightly with Status to return stuff, so I may create a covered index that uses these two, but stores status too

src/SFA.DAS.ApplyService.Domain/Apply/Answer.cs
MarkFCain: That sholudn't have been changed, I must have missed it in my final review before TRing
.  I did a replace value with status in my object, and it must have picked this up.  I'm changing that back.

src/SFA.DAS.ApplyService.Database/Tables/GatewayAnswer.sql
MarkFCain: index now added

src/SFA.DAS.ApplyService.Domain/Apply/Answer.cs
MarkFCain: changed back now

src/SFA.DAS.ApplyService.Database/SFA.DAS.ApplyService.Database.refactorlog
```diff
@@ -0,0 +1,10 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
```
martin-hornsby: delete this file

PR #490, opened 2020-03-20
APR-1444 add question data API endpoint for criminal compliance pages

David-B-Read:


src/SFA.DAS.ApplyService.InternalApi/Controllers/RoatpGatewayCriminalComplianceChecksController.cs
```diff
@@ -0,0 +1,67 @@
...
+                            if (furtherInformationAnswer != null)
+                            {
+                                criminalComplianceCheckDetails.FurtherQuestionId = furtherInformationAnswer.QuestionId;
+                                criminalComplianceCheckDetails.FurtherAnswer = furtherInformationAnswer.Value;
+                                break;
+                            }
+                        }
+                    }
+                }
+            }
```
MichaelYoung1981: I think this block would benefit from a little bit of refactoring, maybe just extract a few methods to make it obvious at a glance what is going on?

src/SFA.DAS.ApplyService.InternalApi/Services/CriminalComplianceChecksQuestionLookupService.cs
```diff
@@ -0,0 +1,27 @@
...
+using SFA.DAS.ApplyService.InternalApi.Models.Roatp;
+using System;
+
+namespace SFA.DAS.ApplyService.InternalApi.Services
+{
+    public class CriminalComplianceChecksQuestionLookupService : ICriminalComplianceChecksQuestionLookupService
+    {
+        public QnaQuestionDetails GetQuestionDetailsForGatewayPageId(string gatewayPageId)
+        {
+            switch (gatewayPageId)
```
MichaelYoung1981: Whenever I see a switch I get nervous about how the code is going to end up.  As there is only 1 path I'd say this doesn't need a switch statement for now so we should remove it and look at alternative patterns as and when it grows.

David-B-Read: Currently only one, yes. The intent is for the same view/view model to be re-used for all the criminal compliance questions as they all have the same format. This is a lookup from gateway id to the QnA data it needs to lookup to populate the page

src/SFA.DAS.ApplyService.InternalApi/Controllers/RoatpGatewayCriminalComplianceChecksController.cs
David-B-Read: Updated so it's clearer what the inner loop is for

PR #494, opened 2020-03-23
Fix lifting logic from admin to apply

MarkFCain:


src/SFA.DAS.ApplyService.InternalApi/Controllers/GatewaExperienceAndAccreditationController.cs
```diff
@@ -0,0 +1,49 @@
...
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Authorization;
+using Microsoft.AspNetCore.Mvc;
+using SFA.DAS.ApplyService.Application.Apply.Roatp;
+using SFA.DAS.ApplyService.InternalApi.Infrastructure;
+
+namespace SFA.DAS.ApplyService.InternalApi.Controllers
+{
+    [Authorize]
+    public class GatewaExperienceAndAccreditationController : Controller
```
David-B-Read: Typo in controller name

src/SFA.DAS.ApplyService.InternalApi/Controllers/GatewayOrganisationChecksController.cs
```diff
@@ -0,0 +1,49 @@
...
+
+namespace SFA.DAS.ApplyService.InternalApi.Controllers
+{
+    [Authorize]
+    public class GatewayOrganisationChecksController : Controller
+    {
+        private readonly IInternalQnaApiClient _qnaApiClient;
+
+        /// <summary>
+        /// Returns trading name for an application
```
David-B-Read: Should this comment be on the GetTradingName action?

src/SFA.DAS.ApplyService.InternalApi/Controllers/GatewaExperienceAndAccreditationController.cs
MichaelYoung1981: The controller name seems to be specific to the Gateway page it's serving, I'm not sure that's the right way as it stops them being reused.

src/SFA.DAS.ApplyService.InternalApi/Controllers/GatewayOrganisationChecksController.cs
MarkFCain: Hah, when I wrote that, there was onl;y one call, and I forgot to update it.  Now done

src/SFA.DAS.ApplyService.InternalApi/Controllers/GatewaExperienceAndAccreditationController.cs
MichaelYoung1981: It could just be as simple as removing the word Gateway from the controller

src/SFA.DAS.ApplyService.InternalApi/Controllers/GatewayOrganisationChecksController.cs
MichaelYoung1981: Should the comment exist at all? 😁

MichaelYoung1981:
General comment...no unit tests.  We all need to get into the habit of test first.

src/SFA.DAS.ApplyService.Application/Apply/Roatp/RoatpPreambleQuestionBuilder.cs
```diff
@@ -158,7 +163,8 @@ public class DescribeYourOrganisation
 
         public class ExperienceAndAccreditations
         { 
-            public static string StartPage = "235";
+            public static string OfficeForStudents = "235";
+            public static string InititalTeacherTraining = "240";
```
David-B-Read: Typo in const

src/SFA.DAS.ApplyService.InternalApi/Controllers/ExperienceAndAccreditationController.cs
```diff
@@ -0,0 +1,49 @@
...
+
+namespace SFA.DAS.ApplyService.InternalApi.Controllers
+{
+    [Authorize]
+    public class ExperienceAndAccreditationController : Controller
+    {
+        private readonly IInternalQnaApiClient _qnaApiClient;
+
+        /// <summary>
+        /// Returns trading name for an application
```
David-B-Read: Comment is misleading

src/SFA.DAS.ApplyService.InternalApi/Controllers/ExperienceAndAccreditationController.cs
```diff
@@ -0,0 +1,49 @@
...
+
+        [HttpGet("/Gateway/{applicationId}/OfficeForStudent")]
+        public async Task<string> GetOfficeForStudents(Guid applicationId)
+        {
+            var websiteNamePage = await _qnaApiClient.GetPageBySectionNo(applicationId, 
+                RoatpWorkflowSequenceIds.YourOrganisation, 
+                RoatpWorkflowSectionIds.YourOrganisation.ExperienceAndAccreditations, 
+                RoatpWorkflowPageIds.ExperienceAndAccreditations.OfficeForStudents);
+
+            return websiteNamePage?.PageOfAnswers?.SelectMany(a => a.Answers)?
```
David-B-Read: Wondering if we could make this a function in a base class to avoid repeated code across multiple controllers?

MarkFCain: possibly.  I like the idea of a base class when you know it's the same sequence and section, and only the page changes.  Let's leave it as is, but establish a pattern whenever you see more than 2? That sound good?

martin-hornsby: is it a good idea to return a string from a REST API? Normally it would be JSON (hence like the Answer object being used in QnA). Also it limits you somewhat if you need to expand the returned value into a object

MarkFCain: We return stuff like that when it's not a primitive.  It's just these ones are just a string....

We'll have other logical entities if needed.  I think Dave's done stuff like that in Criminal Compliance

src/SFA.DAS.ApplyService.InternalApi.UnitTests/ExperienceAndAccreditationControllerTests.cs
```diff
@@ -0,0 +1,177 @@
...
+                                    RoatpYourOrganisationQuestionIdConstants.OfficeForStudents,
+                                Value = ValueOfQuestion
+                            }
+                        }
+                    }
+                }
+            };
+
+            _qnaApiClient.Setup(x =>
+                    x.GetPageBySectionNo(It.IsAny<Guid>(), It.IsAny<int>(), It.IsAny<int>(), It.IsAny<string>()))
```
MichaelYoung1981: IsAny shouldn't be used, we know the values that we expect.

src/SFA.DAS.ApplyService.InternalApi.UnitTests/ExperienceAndAccreditationControllerTests.cs
```diff
@@ -0,0 +1,177 @@
...
+                .ReturnsAsync(page);
+
+            var actualResult = _controller.GetOfficeForStudents(new Guid()).Result;
+
+            Assert.AreEqual(ValueOfQuestion,actualResult);
+            _qnaApiClient.Verify(x=>x.GetPageBySectionNo(It.IsAny<Guid>(), It.IsAny<int>(), It.IsAny<int>(), It.IsAny<string>()),Times.Once);
```
MichaelYoung1981: This probably isn't necessary (although I understand why people put them in).  If the method wasn't called you wouldn't get the right actualResult.

src/SFA.DAS.ApplyService.InternalApi.UnitTests/ExperienceAndAccreditationControllerTests.cs
```diff
@@ -0,0 +1,177 @@
...
+using SFA.DAS.ApplyService.Domain.Entities;
+using SFA.DAS.ApplyService.Domain.Roatp;
+using SFA.DAS.ApplyService.InternalApi.Types.CompaniesHouse;
+using SFA.DAS.ApplyService.InternalApi.Types;
+using SFA.DAS.ApplyService.InternalApi.Services;
+
+namespace SFA.DAS.ApplyService.InternalApi.UnitTests
+{
+    [TestFixture]
+   public class ExperienceAndAccreditationControllerTests
```
MichaelYoung1981: There seems to be quite a lot of repetition in the class.  Could some of it be moved the SetUp or abstracted out into methods that are reused?

src/SFA.DAS.ApplyService.InternalApi/Controllers/GatewayOrganisationChecksController.cs
```diff
@@ -0,0 +1,45 @@
...
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Authorization;
+using Microsoft.AspNetCore.Mvc;
+using SFA.DAS.ApplyService.Application.Apply.Roatp;
+using SFA.DAS.ApplyService.InternalApi.Infrastructure;
+
+namespace SFA.DAS.ApplyService.InternalApi.Controllers
+{
+    [Authorize]
+    public class GatewayOrganisationChecksController : Controller
```
MichaelYoung1981: Are there tests for this code?

MarkFCain: Seems not, I thought there were.  I'll put some in.

PR #497, opened 2020-03-25
APR-1477 RoATP FHA functional & design update

martin-hornsby:


src/SFA.DAS.ApplyService.Domain/Apply/RoaptFinancialApplicationsStatusCounts.cs
```diff
@@ -0,0 +1,9 @@
+﻿namespace SFA.DAS.ApplyService.Domain.Apply
+{
+    public class RoaptFinancialApplicationsStatusCounts
```
David-B-Read: Should be RoatpFinancialApplicationsStatusCounts

src/SFA.DAS.ApplyService.Domain/Apply/RoaptFinancialApplicationsStatusCounts.cs
```diff
@@ -0,0 +1,9 @@
...
+    {
+        public int ApplicationsOpen { get; set; }
+        public int ApplicationsWithClairification { get; set; }
```
David-B-Read: Typo in property name

src/SFA.DAS.ApplyService.Data/ApplyRepository.cs
```diff
@@ -478,6 +478,21 @@ FROM OPENJSON(i.Sequences)
             }
         }
 
+        public async Task<RoaptFinancialApplicationsStatusCounts> GetFinancialApplicationsStatusCounts()
+        {
+            // Note: For now it is easier to run all three queries. It may make sense to do something similar to that done with EPAO
```
David-B-Read: Should be ok for Private Beta, but we might want to look at just getting the counts once more applications start to come through

src/SFA.DAS.ApplyService.Domain/Apply/RoaptFinancialApplicationsStatusCounts.cs
martin-hornsby: corrected

src/SFA.DAS.ApplyService.Domain/Apply/RoaptFinancialApplicationsStatusCounts.cs
martin-hornsby: corrected

PR #500, opened 2020-03-26
Apr 1337 subcontractor declaration

MichaelYoung1981:


src/SFA.DAS.ApplyService.InternalApi/Infrastructure/InternalQnaApiClient.cs
```diff
@@ -12,7 +14,7 @@
 
 namespace SFA.DAS.ApplyService.InternalApi.Infrastructure
 {
-    public class InternalQnaApiClient:IInternalQnaApiClient
+    public class InternalQnaApiClient : IInternalQnaApiClient
```
David-B-Read: As we're adding more and more functionality into the QnA API Client used by the Internal API, would it be better to share the one thats in ApplyService.Web in an Api Clients project?

MichaelYoung1981: Good question.  There is currently only one common method between the two classes.  Hows about the next time someone adds a method that already exists on the one in ApplyService.Web we get them to abstract out the common methods? 😁

David-B-Read: GetPageBySectionNo, GetQuestionTag, GetAnswerValue, DownloadFile are all implemented in the ApplyService.Web client.

MarkFCain: Yeah the ones in this are in both.  I don't know whether a common catchall one is better cos internal doesn't need all the various calls in the web version.  It's more like take the internal one, make it higher level (own project), and remove the same from the web one and repoint any usages of those methods to the higher level one.. I think this makes sense.

David-B-Read:
Merged in https://github.com/SkillsFundingAgency/das-apply-service/pull/529

PR #501, opened 2020-03-26
APR-1322 Website address

MakCom:


src/SFA.DAS.ApplyService.InternalApi.UnitTests/GatewayOrganisationChecksControllerTests.cs
```diff
@@ -117,5 +120,21 @@ public void get_website_address_manually_entered_returns_no_value_when_not_prese
 
             Assert.IsNull(actualResult);
         }
+
+
+        [Test]
```
David-B-Read: This appears to only cover one of the three possible scenarios when looking up the website address

MakCom: One more test scenario added

David-B-Read: The code is handling/logging a message when no website is found - that was the third scenario I was referring to

MakCom: All scenarios now covered.

src/SFA.DAS.ApplyService.InternalApi.UnitTests/GatewayOrganisationChecksControllerTests.cs
```diff
@@ -117,5 +120,45 @@ public void get_website_address_manually_entered_returns_no_value_when_not_prese
...
+               .Setup(x => x.GetAnswerValue(_applicationId,
+                                           RoatpWorkflowSequenceIds.YourOrganisation,
+                                           RoatpWorkflowSectionIds.YourOrganisation.WhatYouWillNeed,
+                                           RoatpWorkflowPageIds.WebsiteManuallyEntered,
+                                           RoatpYourOrganisationQuestionIdConstants.WebsiteManuallyEntered)).ReturnsAsync(applyWebsite);    
+
+
+            var actualResult = _controller.GetOrganisationWebsiteAddress(_applicationId).Result;
+
+            switch (testCase)
```
MichaelYoung1981: This definitely shouldn't be here.  If we are checking the case number we should have 4 different tests.

MichaelYoung1981: Obviously reuse all of the setup above tho, no need to copy it all to each test.

MakCom: Tests are refactored.

PR #502, opened 2020-03-30
Amended initial teacher training endpoint to return additional question.

MichaelYoung1981:


src/SFA.DAS.ApplyService.Application/Apply/Roatp/RoatpPreambleQuestionBuilder.cs
```diff
@@ -172,6 +173,7 @@ public class ExperienceAndAccreditations
         { 
             public static string OfficeForStudents = "235";
             public static string InitialTeacherTraining = "240";
+            public static string IsPostGradTrainingOnlyApprenticeship = "240";
```
David-B-Read: This should be page Id 250

MichaelYoung1981: Corrected. 👍

PR #510, opened 2020-04-02
Check the organisation type to determine which question should have b…

MichaelYoung1981:
…een answered.


src/SFA.DAS.ApplyService.Web/Controllers/Roatp/RoatpApplicationController.cs
```diff
@@ -514,16 +514,24 @@ public async Task<IActionResult> TaskList(Guid applicationId)
 
             var whosInControlConfirmed = false;
 
-            var soleTraderDateOfBirthAnswer = await _qnaApiClient.GetAnswer(applicationId, whosInControlSection.Id, RoatpWorkflowPageIds.WhosInControl.AddSoleTraderDob, RoatpYourOrganisationQuestionIdConstants.AddSoleTradeDob);
-            if (soleTraderDateOfBirthAnswer != null && !String.IsNullOrEmpty(soleTraderDateOfBirthAnswer.Value))
+            var organisationType = await _qnaApiClient.GetAnswer(applicationId, whosInControlSection.Id, RoatpWorkflowPageIds.WhosInControl.SoleTraderPartnership, RoatpYourOrganisationQuestionIdConstants.SoleTradeOrPartnership);
+            if (organisationType != null && organisationType.Value == "Sole trader")
```
David-B-Read, line 518: Can this, and Partnership below, be put into a constants file?

src/SFA.DAS.ApplyService.Web.UnitTests/TaskListViewModelTests.cs
```diff
@@ -628,10 +628,26 @@ public void Whos_in_control_section_status_shows_as_completed_if_companies_house
...
+        {
+            var model = GetTaskListViewModelWithSectionsUpToWhosInControlCompleted();
+
+            model.VerifiedCompaniesHouse = false;
+            model.VerifiedCharityCommission = false;
+            model.CompaniesHouseDataConfirmed = false;
+            model.CharityCommissionDataConfirmed = false;
+            model.WhosInControlConfirmed = false;
+            model.WhosInControlStarted = true;
+
```
MarkFCain, line 647: you only need WhoIsInControlStarted = true for the Should to be true.  Either take out the rest, or put in test cases where they all vary to check the Should() isn't affected.

PR #512, opened 2020-04-02
Apr 1293 2 1 people in control

MarkFCain:


src/SFA.DAS.ApplyService.InternalApi.Types/GatewayOrganisationTypes.cs
```diff
@@ -0,0 +1,16 @@
...
+namespace SFA.DAS.ApplyService.InternalApi.Types
+{
+    public class GatewayOrganisationTypes
+    {
+        public static string Company = "Company";
+        public static string CompanyAndCharity = "Company and charity";
+        public static string Charity = "Charity";
+        public static string SoleTrader = "Sole trader";
+        public static string Partnership = "Partnership";
+        public static string StatutoryInstitute = "Statutory institute";
```
MichaelYoung1981, line 14: These should be consts

src/SFA.DAS.ApplyService.Application.UnitTests/Services/DateOfBirthFormatterTests.cs
```diff
@@ -66,5 +66,16 @@ public void Format_date_of_birth_rejects_non_numeric_values(string month, string
 
             formattedDob.Should().Be(expectedValue);
         }
+
+        [TestCase("1,1980", "Jan 1980")]
+        [TestCase("2,1981", "Feb 1981")]
+        [TestCase("1981", "")]
+        [TestCase("1", "")]
+        [TestCase("13,1980", "")]
+        public void Format_monthNumber_comma_year_converted_to_mmm_yyyy(string monthYearDescription, string expectedValue)
```
MichaelYoung1981, line 75: I think I'd rather the positive and negative cases were separate tests.

src/SFA.DAS.ApplyService.Domain/Entities/PersonInControl.cs
```diff
@@ -0,0 +1,9 @@
+﻿namespace SFA.DAS.ApplyService.Domain.Entities
+{ 
+
+public class PersonInControl
+    {
+        public string Name { get; set; }
+        public string DateOfBirth { get; set; }
```
MichaelYoung1981, line 7: Rather than have a formatter do you think it'd be worth having a DateOfBirth value type?

src/SFA.DAS.ApplyService.InternalApi.UnitTests/OrganisationSummaryControllerGetCompanyDirectorTests.cs
```diff
@@ -0,0 +1,143 @@
...
+            _applyRepository.Setup(x => x.GetApplyData(It.IsAny<Guid>())).ReturnsAsync((ApplyData) null);
+        }
+
+
+        [Test]
+        public void get_no_directors_if_tag_doesnt_exist()
+        {
+            var result = (OkObjectResult)_controller.GetDirectorsFromSubmitted(_applicationId).Result;
+
+            result.Should().BeOfType<OkObjectResult>();
```
MichaelYoung1981, line 46: It's not possible for this assertion to ever fail given the cast above is it?

src/SFA.DAS.ApplyService.InternalApi/Controllers/OrganisationSummaryController.cs
```diff
@@ -0,0 +1,296 @@
...
+            _logger = logger;
+        } 
+
+        [HttpGet]
+        [Route("TypeOfOrganisation/{applicationId}")]
+        public async Task<IActionResult> GetTypeOfOrganisation(Guid applicationId)
+        {
+            _logger.LogInformation($"Retrieving type of organisation for application {applicationId}");
+            const string TRUE = "TRUE";
+                 var companyVerification = await
```
MichaelYoung1981, line 38: Formatting looks a touch out.

src/SFA.DAS.ApplyService.InternalApi/Controllers/OrganisationSummaryController.cs
```diff
@@ -0,0 +1,296 @@
...
+                await _qnaApiClient.GetTabularDataByTag(applicationId,
+                    RoatpWorkflowQuestionTags.CompaniesHouseDirectors);
+
+            if (directorsData?.DataRows == null || !directorsData.DataRows.Any()) return Ok(peopleInControl);
+
+            foreach (var director in directorsData.DataRows.Where(x => x.Columns.Any()).OrderBy(x => x.Columns[0]))
+            {
+                var directorName = director.Columns[0];
+                var directorDob = string.Empty;
+                if (director.Columns.Any() && director.Columns.Count >= 2)
```
MichaelYoung1981, line 86: If there aren't any columns line 84 would fail so not sure if this part of the check is necessary (or it could be in the wrong place?)

src/SFA.DAS.ApplyService.InternalApi/Controllers/OrganisationSummaryController.cs
```diff
@@ -0,0 +1,296 @@
...
+                {
+                    var picName = pic.Columns[0];
+                    var picDob = string.Empty;
+                    if (pic.Columns.Any() && pic.Columns.Count >= 2)
+                    {
+                        picDob = pic.Columns[1];
+                    }
+
+                    peopleInControl.Add(new PersonInControl { Name = picName, DateOfBirth = picDob });
+                }
```
MichaelYoung1981, line 278: Can the contents of this if be extracted out into a private? Same for the else block below.

src/SFA.DAS.ApplyService.InternalApi.Types/GatewayOrganisationTypes.cs
```diff
@@ -0,0 +1,16 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace SFA.DAS.ApplyService.InternalApi.Types
+{
+    public class GatewayOrganisationTypes
```
MichaelYoung1981, line 7: Could you rename the class to RoatpOrganisationTypes?  I've created the same class on a branch off master so we need to make sure we don't get two classes with the same constants.

src/SFA.DAS.ApplyService.Domain/Entities/PersonInControl.cs
MarkFCain: This conforms to the common output of MMM yyyy instead of d/m/y type of stuff, so it would still be a string.  Maybe the words MonthYearOfBirth would be a better name though?

src/SFA.DAS.ApplyService.InternalApi/Controllers/OrganisationSummaryController.cs
MarkFCain: There's a Where clause a few lines above that filters on Columns.Any()

MichaelYoung1981: There are but 2 lines up would still fail if there are no columns wouldn't it? var directorName = director.Columns[0];

MarkFCain: There's an any in the foreach that checks there's always at least one

MarkFCain: Hi Mike, I can't seem to get a re-review requested for this in the github, and as you're sick and I've been asked to merge quite a few bits in, I'm going ahead and resolving this conversation so I can merge it.  I have double checked it all addresses your concerns though.

PR #513, opened 2020-04-06
Gateway submaster

MarkFCain:


src/SFA.DAS.ApplyService.Domain/Roatp/GatewayPageIds.cs
```diff
@@ -0,0 +1,25 @@
...
+
+        public const string WebsiteAddress = "WebsiteAddress";
+
+
+        public const string OfficeForStudent = "OfficeForStudents";
+        public const string InitialTeacherTraining = "InitialTeacherTraining";
+        public const string Ofsted = "Ofsted";
+        public const string SubcontractorDeclaration = "SubcontractorDeclaration";
+
+        public const string CCOrganisationCompositionCreditors = "CompositionWithCreditors";
```
David-B-Read, line 21: This looks out of date, there should be subclasses in here for criminal compliance page Ids

David-B-Read, line 21: These criminal compliance page ids can be removed at some point, they're referenced in config now

PR #518, opened 2020-04-08
APR-1570 - Add error logging to API clients 

martin-hornsby:


src/SFA.DAS.ApplyService.InternalApi/Infrastructure/ApiClientBase.cs
```diff
@@ -0,0 +1,232 @@
+﻿using Microsoft.Extensions.Logging;
+using Newtonsoft.Json;
+using System;
+using System.Net;
+using System.Net.Http;
+using System.Threading.Tasks;
+
+namespace SFA.DAS.ApplyService.InternalApi.Infrastructure
+{
+    public abstract class ApiClientBase<CB>
```
MichaelYoung1981, line 10: This class should be reused, it seems there are 2 copies of it?

martin-hornsby: I can try to figure out if we can shove this somewhere else. Seems overkill to have a new class library for it. I'd imagine sticking it anywhere else due to it being Infrastructure would create problems though.

MichaelYoung1981: If we end up with a new project I'm sure there are other pieces of duplicate code which could go in it.

src/SFA.DAS.ApplyService.InternalApi/Infrastructure/AssessorServiceApiClient.cs
```diff
@@ -16,27 +16,28 @@
...
+            }
+            _httpClient.DefaultRequestHeaders.Authorization = new AuthenticationHeaderValue("Bearer", GetToken());
         }
 
         public async Task<IEnumerable<Types.OrganisationSearchResult>> SearchOrgansiation(string searchTerm)
         {
             _logger.LogInformation($"Searching EPAO Register. Search Term: {searchTerm}");
             var apiResponse =
-                await Get<IEnumerable<OrganisationSummary>>($"/api/ao/assessment-organisations/search/{searchTerm}");
+                await Get<List<OrganisationSummary>>($"/api/ao/assessment-organisations/search/{searchTerm}");
```
MarkFCain, line 40: There's a lot of this replacing IEnumerable with List, and the reason isn;'t clear in the code.  What's it for?

martin-hornsby: Needed for generics to work correctly

MarkFCain: ok

src/SFA.DAS.ApplyService.InternalApi/Infrastructure/ApiClientBase.cs
martin-hornsby: Done

src/SFA.DAS.ApplyService.InternalApi/Startup.cs
```diff
@@ -215,7 +175,56 @@ public void Configure(IApplicationBuilder app, IHostingEnvironment env)
...
         }
-        
+
+        private void ConfigHttpClients(IServiceCollection services)
+        {
+            var acceptHeaderName = "Accept";
+            var acceptHeaderValue = "application/json";
+            var handlerLifeTime = TimeSpan.FromMinutes(5);
+
+            services.AddHttpClient<AssessorServiceApiClient, AssessorServiceApiClient>(config =>
```
David-B-Read, line 185: Some of these clients are EPAO-related, are they needed in Apply now it's RoATP-centric?

martin-hornsby: It should be used for RoEPAO checks only for gateway 😺

PR #522, opened 2020-04-14
APR-1576 - Add Polly to ApiClient

martin-hornsby:


David-B-Read:
> Can the number of retries be put in config?

Would make sense for the retry delay to go in there as well

martin-hornsby:
Further investigation needs to be done on this branch as it appears Polly is not throwing exceptions up the stack in some cases and returns null instead

PR #524, opened 2020-04-15
APR-977 - skip button styling

David-B-Read:


src/SFA.DAS.ApplyService.Web/Views/Application/Pages/Index.cshtml
```diff
@@ -271,28 +271,16 @@
...
-                            if (first && Model.SectionId != 1)
-                            {
-
-                                <a class="govuk-link das-inline-button-link govuk-link--no-visited-state" asp-action="@link.Action" asp-controller="@link.Controller" asp-route-applicationId="@Model.ApplicationId" asp-route-sequenceId="@Model.SequenceId" asp-route-sectionId="@Model.SectionId" asp-route-pageId="@pageId">@link.Caption</a>
-
-                            }
-                            else
-                            {
-                                first = false;
-                                <p class="govuk-body">
```
gdbarnes, line 292: This markup is correct for the link below the button (the paragraph wrapping the link)

We shouldn't be forcing links to be block elements using css. One good reason is that the clickable area will span the width of the containing box rather than the text.

src/SFA.DAS.ApplyService.Web/Views/Application/Pages/Index.cshtml
```diff
@@ -271,28 +271,16 @@
...
-                                    <a class="govuk-link" asp-action="@link.Action" asp-controller="@link.Controller" asp-route-applicationId="@Model.ApplicationId" asp-route-sequenceId="@Model.SequenceId" asp-route-sectionId="@Model.SectionId" asp-route-pageId="@pageId">@link.Caption</a>
-                                </p>
```
gdbarnes, line 294: This existing markup is correct (the paragraph tag wrapping the link tag) for the link below the button.

We shouldn't be forcing links to be block elements using css.

One good reason is that the clickable area will span the width of the containing box rather than the text.

David-B-Read: Reverted back to the original logic

gdbarnes: 👍

PR #526, opened 2020-04-16
APR-1311 throw service unavailable exceptions if external calls fail

David-B-Read:


martin-hornsby:
1570 has been merged into master

PR #527, opened 2020-04-17
APR-1362 - govuk frontend upgrade

gdbarnes:
This branch must be tested and deployed at the same time as:

`APR-1362_govuk_frontend_upgrade` on `das-assessor-service`

and

`APR-1362_govuk_frontend_upgrade` on `das-qna-api`


src/SFA.DAS.ApplyService.Web/Views/Application/PagesWithSections.cshtml
```diff
@@ -25,25 +25,23 @@
...
                                 isSamePage = pageTitle == pg.Title;
                                 pageTitle = pg.Title;
                                 nextPageTitle = "";
+
                                 if (i < pagesToCheck.Count - 1)
                                 {
                                     nextPageTitle = (pagesToCheck[i + 1]).Title;
                                 }
 
-                                @if (!isSamePage)
```
MarkFCain, line 42: are you sure this should go?

src/SFA.DAS.ApplyService.Web/Views/Application/PagesWithSections.cshtml
```diff
@@ -72,17 +71,19 @@
 
                                 @if (nextPageTitle != pageTitle)
                                 {
-                                    @Html.Raw("</ul>");
+                                    @Html.Raw("</ul></li>");
```
MarkFCain, line 74: this </li> bit looks wrong to me

MarkFCain (approved):

src/SFA.DAS.ApplyService.Web/Views/Application/PagesWithSections.cshtml
gdbarnes: I'm pretty sure that's right. I know what you're getting at, but it's closing a list item which contains an unordered list so it looks the wrong way round.

src/SFA.DAS.ApplyService.Web/Views/Application/PagesWithSections.cshtml
```diff
@@ -25,25 +25,23 @@
             else
             {
                 <h1 class="govuk-heading-xl">@Model.Title</h1>
-                <ul class="govuk-list">
+
+                <ol class="app-task-list">
                     <li>
-                        <ul class="app-task-list govuk-!-margin-bottom-7">
+                        <h2 class="app-task-list__section">@Html.Raw(pagesToCheck[0].Title)</h2>
+                        <ul class="app-task-list__items app-task-list__items--no-number">
```
martin-hornsby, line 32: Have we checked all of this displays correctly as it's feels like very complex logic (which could probably benefit from being re-worked). From what I understand it's looking through all pages to check, and when it is a 'new page (i.e. a new title)' it dumps out the new title and puts stuff into a new unordered list 🤷‍♂️

gdbarnes: Seems to display correctly but I'm happy for the logic to be re-worked at a later date.

src/SFA.DAS.ApplyService.Web/Views/Application/PagesWithSections.cshtml
gdbarnes: I've removed all reference to that now. I can't see any issue with the UI on first glance but lets see what comes out of testing.

MarkFCain (approved):

martin-hornsby (approved):

PR #528, opened 2020-04-20
APR-1492 About to confirm gateway outcome - Pass

MakCom:
APR-1492 About to confirm gateway outcome - Pass
&
APR-1485 Gateway outcome confirmation


src/SFA.DAS.ApplyService.Domain/Entities/Apply.cs
```diff
@@ -127,8 +127,9 @@ public static class GatewayReviewStatus
         public const string Draft = "Draft";
         public const string New = "New";
         public const string InProgress = "In Progress";
-        public const string Approved = "Approved";
-        public const string Declined = "Declined";
+        public const string Pass = "Pass";
+        public const string Decline = "Decline";
+        public const string AskForClarification = "Ask for clarification";
```
David-B-Read, line 132: Gateway review status field is 20 characters in length, this will be truncated unless the schema is updated

MakCom: I have increased it to 50 characters (look in src/SFA.DAS.ApplyService.Database/Tables/Apply.sql file), so it will not get truncated. But it is also possible "Ask for clarification" to be changed to "Clarification sent" after our Statuses meeting.

PR #530, opened 2020-04-23
APR-1024 Qna View Partials to handle Page Caption

MakCom:


src/SFA.DAS.ApplyService.Web/Controllers/Roatp/RoatpApplicationController.cs
```diff
@@ -318,22 +320,26 @@ public async Task<IActionResult> Page(Guid applicationId, int sequenceId, int se
...
-                if (page == null)
-                {
-                    return RedirectToAction("TaskList", new {applicationId = applicationId});
-                }
-
-                var section = await _qnaApiClient.GetSection(applicationId, selectedSection.Id);
+                //if (page == null)
+                //{
+                //    return RedirectToAction("TaskList", new {applicationId = applicationId});
+                //}
```
MichaelYoung1981, line 340: Can this be deleted rather than commented out?

martin-lush:
Decided we don't use this functionality in RoATP

PR #533, opened 2020-04-28
APR-1490 : Assessor Overview

martin-hornsby:


src/SFA.DAS.ApplyService.InternalApi/Controllers/RoatpAssessorController.cs
```diff
@@ -0,0 +1,62 @@
...
+                GetAssessorSequence(allQnaSections, RoatpWorkflowSequenceIds.ReadinessToEngage, "Readiness to engage checks"),
+                GetAssessorSequence(allQnaSections, RoatpWorkflowSequenceIds.PlanningApprenticeshipTraining, "Planning apprenticeship training checks"),
+                GetAssessorSequence(allQnaSections, RoatpWorkflowSequenceIds.DeliveringApprenticeshipTraining, "Delivering apprenticeship training checks"),
+                GetAssessorSequence(allQnaSections, RoatpWorkflowSequenceIds.EvaluatingApprenticeshipTraining, "Evaluating apprenticeship training checks")
+            };
+        }
+
+        private AssessorSequence GetAssessorSequence(IEnumerable<ApplicationSection> qnaSections, int sequenceNumber, string sequenceTitle)
+        {
+            const int introductionSectionNumber = 1;
```
MarkFCain, line 48: I know all introductions are section number 1, but can you put it in some CONST somewhere and use that?

src/SFA.DAS.ApplyService.InternalApi/Infrastructure/InternalQnaApiClient.cs
```diff
@@ -43,6 +43,22 @@ public InternalQnaApiClient(IConfigurationService configurationService, ILogger<
 
             _environmentName = configurationService.GetEnvironmentName();
         }
+
+        public async Task<IEnumerable<ApplicationSequence>> GetSequences(Guid applicationId)
```
David-B-Read, line 47: This method doesn't seem to be used, is it needed when GetSections returns all sections by sequence?

martin-hornsby: I added it for convenience thinking I would need it, however it turns out I don't so have removed it :)

src/SFA.DAS.ApplyService.InternalApi/Controllers/RoatpAssessorController.cs
martin-hornsby: Done. Added a function and expanded the list of known sequences and sections

src/SFA.DAS.ApplyService.Application/Apply/Roatp/RoatpPreambleQuestionBuilder.cs
```diff
@@ -123,28 +123,88 @@ public static class YourOrganisation
...
-            public static int CommercialInConfidenceInformation = 2;
-            public static int TermsAndConditions = 3;
-            public static int SubmitApplication = 4;
+            public const int ApplicationPermissionsAndChecks = 1;
+            public const int CommercialInConfidenceInformation = 2;
+            public const int TermsAndConditions = 3;
+            public const int SubmitApplication = 4;
+        }
+
+        public static List<int> GetWhatYouWillNeedSectionsForSequence(int sequenceId)
```
David-B-Read, line 172: This probably belongs in a separate class / file. Wondered if it would be better modelled in configuration?

src/SFA.DAS.ApplyService.InternalApi/Controllers/RoatpAssessorController.cs
```diff
@@ -48,6 +61,56 @@ public async Task<List<RoatpAssessorApplicationSummary>> InProgressApplications(
...
+                SequenceTitle = GetSequenceTitle(sequenceNumber),
+                Sections = qnaSections.Where(sec => sec.SequenceId == sequenceNumber && !sectionsToExclude.Contains(sec.SectionId))
+                .Select(sec =>
+                {
+                    return new AssessorSection { SectionNumber = sec.SectionId, LinkTitle = sec.Title, Status = string.Empty };
+                }).ToList()
+            };
+        }
+
+        private static string GetSequenceTitle(int sequenceId)
```
David-B-Read, line 96: Would this be better as a lookup against appsettings?

PR #536, opened 2020-04-30
Apr 1568 oversight tab endpoints

MarkFCain:


src/SFA.DAS.ApplyService.Data/ApplyRepository.cs
```diff
@@ -1130,6 +1130,55 @@ public async Task RemoveSectionCompleted(Guid applicationId, Guid applicationSec
...
+					        JSON_VALUE(apply.ApplyData, '$.ApplyDetails.UKPRN') AS Ukprn,
+                            REPLACE(JSON_VALUE(apply.ApplyData, '$.ApplyDetails.ProviderRouteName'),' provider','') AS ProviderRoute,
+							JSON_VALUE(apply.ApplyData, '$.ApplyDetails.ReferenceNumber') AS ApplicationReferenceNumber,
+                            JSON_VALUE(apply.ApplyData, '$.ApplyDetails.ApplicationSubmittedOn') AS ApplicationSubmittedDate,
+							apply.OversightStatus,
+							Apply.ApplicationDeterminedDate
+                              FROM Apply apply
+	                      INNER JOIN Organisations org ON org.Id = apply.OrganisationId
+	                      WHERE apply.DeletedAt IS NULL
+                          and GatewayReviewStatus  in ('Pass','Fail')
```
David-B-Read, line 1150: This would have to be Pass for the other statuses to be activated

src/SFA.DAS.ApplyService.Data/ApplyRepository.cs
```diff
@@ -1130,6 +1130,55 @@ public async Task RemoveSectionCompleted(Guid applicationId, Guid applicationSec
...
```
David-B-Read, line 1175: Same comment as above

src/SFA.DAS.ApplyService.Data/ApplyRepository.cs
MarkFCain: Yes, it was a pass originally, but the state diagram shows it flows to oversight even when it's a fail.... I'll change it to pass anyway, we all know it actually should never appear if gateway is a fail

PR #540, opened 2020-05-01
Apr 1530 assessor in progress tab

MichaelYoung1981:


src/SFA.DAS.ApplyService.Application.UnitTests/Handlers/InProgressAssessorApplicationsHandlerTests/InProgressAssessorApplicationsHandlerTests.cs
```diff
@@ -0,0 +1,37 @@
...
+            _repository = new Mock<IAssessorRepository>();
+            _handler = new InProgressAssessorApplicationsHandler(_repository.Object);
+        }
+
+        [Test]
+        public async Task Get_in_progress_applications_returns_in_progress_applications_for_the_user()
+        {
+            var expectedUser = "sadjkffgdji";
+            var expectedResult = new List<RoatpAssessorApplicationSummary>();
+            _repository.Setup(x => x.GetInProgressAssessorApplications(expectedUser)).ReturnsAsync(expectedResult);
```
MarkFCain, line 30: Can you add at least one item to this list and test it's in the actual?

MichaelYoung1981: It's comparing the instance returned from the method is the same as what is returned from the mock, adding an item to the list won't make the blindest bit of difference.

PR #544, opened 2020-05-05
APR-1680 fix Apply API exception if ITT answer is No

David-B-Read:


src/SFA.DAS.ApplyService.InternalApi/Controllers/ExperienceAndAccreditationController.cs
```diff
@@ -32,24 +32,31 @@ public async Task<string> GetOfficeForStudents(Guid applicationId)
...
+                    RoatpWorkflowSectionIds.YourOrganisation.ExperienceAndAccreditations,
+                    RoatpWorkflowPageIds.ExperienceAndAccreditations.IsPostGradTrainingOnlyApprenticeship,
+                    RoatpYourOrganisationQuestionIdConstants.IsPostGradTrainingOnlyApprenticeship);
+
+                return new InitialTeacherTraining
+                {
+                    DoesOrganisationOfferInitialTeacherTraining = true,
+                    IsPostGradOnlyApprenticeship = isPostGradTrainingOnlyApprenticeship.ToUpper() == "YES"
+                };
+            }
```
MichaelYoung1981, line 54: Would it have been easier to just change isPostGradTrainingOnlyApprenticeshipTask.Result.ToUpper() == "YES" to isPostGradTrainingOnlyApprenticeshipTask.Result?.ToUpper() == "YES"

David-B-Read: No, because the QnA API responds with a 404 error when the question doesn't have an answer.

MichaelYoung1981: Gotcha.  Is that 404 consistent?  Lots of the other methods use the null check on the result and they seem to be working?

David-B-Read: It's worth checking, changes to logging/exception handling mean that anything that's not a successful response is thrown

PR #546, opened 2020-05-06
Apr 1569 oversight overall outcome

MarkFCain:


src/SFA.DAS.ApplyService.InternalApi/Controllers/OversightController.cs
```diff
@@ -1,8 +1,10 @@
-﻿using System.Collections.Generic;
+﻿using System;
+using System.Collections.Generic;
 using System.Threading.Tasks;
 using MediatR;
 using Microsoft.AspNetCore.Authorization;
 using Microsoft.AspNetCore.Mvc;
+using NPOI.HPSF;
```
David-B-Read, line 7: Is this needed?

src/SFA.DAS.ApplyService.Application.UnitTests/Handlers/OversightHandlerTests/OversightDetailsHandlerTests.cs
```diff
@@ -0,0 +1,53 @@
...
+                ApplicationReferenceNumber = "APR000111",
+                OversightStatus = "Successful",
+                ApplicationDeterminedDate = DateTime.Today
+            };
+
+            ApplyRepository.Setup(r => r.GetOversightDetails(applicationId)).ReturnsAsync(applicationDetails);
+
+            var result = _handler.Handle(new GetOversightDetailsRequest(applicationId), new CancellationToken()).GetAwaiter().GetResult();
+
+            Assert.AreEqual(applicationId, result.ApplicationId);
```
MichaelYoung1981, line 49: This assertion doesn't feel sufficient, an IsSame on the result and applicationDetails would be a better check.

src/SFA.DAS.ApplyService.InternalApi.UnitTests/OversightControllerTests.cs
```diff
@@ -97,5 +97,40 @@ public async Task Check_pending_results_are_as_expected()
...
+
+            _mediator
+                .Setup(x => x.Send(It.IsAny<GetOversightDetailsRequest>(), It.IsAny<CancellationToken>()))
+                .ReturnsAsync(oversight);
+
+            var actualResult = await _controller.OversightDetails(applicationId);
+
+            var returnedOversight = actualResult.Value;
+
+            Assert.AreEqual(oversight.Id, returnedOversight.Id);
```
MichaelYoung1981, line 126: In contract to my previous comment, these checks are probably overkill.  IsSame would be better and more appropriate to the code in the controller.

src/SFA.DAS.ApplyService.Application.UnitTests/Handlers/OversightHandlerTests/OversightDetailsHandlerTests.cs
MarkFCain: Mike I've done the changes but the GIT has got to a state it won't allow me to re-review.  Can I mark your comments as resolved and see if allows things to move on?

PR #551, opened 2020-05-12
APR-1656 - Display pages in RoATP Assessor

martin-hornsby:


src/SFA.DAS.ApplyService.InternalApi/Services/AssessorLookupService.cs
```diff
@@ -0,0 +1,128 @@
...
+namespace SFA.DAS.ApplyService.InternalApi.Services
+{
+    public class AssessorLookupService : IAssessorLookupService
+    {
+        public string GetTitleForSequence(int sequenceId)
+        {
+            switch (sequenceId)
+            {
+                case RoatpWorkflowSequenceIds.ProtectingYourApprentices:
+                    return "Protecting your apprentices checks";
```
MarkFCain, line 12: I'd turn these into public strings, and return the public strings here (either as fields on this class, or probably better to create a separate class and reference it?)

src/SFA.DAS.ApplyService.InternalApi/Controllers/RoatpAssessorController.cs
```diff
@@ -131,23 +134,71 @@ private static List<int> GetWhatYouWillNeedSectionsForSequence(int sequenceNumbe
...
+
+        [HttpGet("Assessor/Applications/{applicationId}/Sequences/{sequenceNumber}/Sections/{sectionNumber}/Page/{pageId}")]
+        public async Task<AssessorPage> GetAssessorPage(Guid applicationId, int sequenceNumber, int sectionNumber, string pageId)
+        {
+            AssessorPage page = null;
+
+            var qnaSection = await _qnaApiClient.GetSectionBySectionNo(applicationId, sequenceNumber, sectionNumber);
+            var qnaPage = qnaSection?.QnAData.Pages.FirstOrDefault(p => p.PageId == pageId || string.IsNullOrEmpty(pageId));
+
+            if (qnaPage != null)
```
David-B-Read, line 152: Is it worth moving this code block into a service or orchestrator?

src/SFA.DAS.ApplyService.InternalApi/Services/AssessorLookupService.cs
```diff
@@ -0,0 +1,128 @@
...
+                case RoatpWorkflowSequenceIds.EvaluatingApprenticeshipTraining:
+                    return "Evaluating apprenticeship training checks";
+                default:
+                    return null;
+            }
+        }
+
+        public string GetTitleForPage(string pageId)
+        {
+            switch (pageId)
```
David-B-Read, line 28: Personally, I'd prefer to put this into an appsettings section rather than a case statement

martin-hornsby: Makes sense if these titles are susceptible to change but they are static for now

src/SFA.DAS.ApplyService.InternalApi/Controllers/RoatpAssessorController.cs
martin-hornsby: This block of code has been refactored

src/SFA.DAS.ApplyService.InternalApi/Controllers/RoatpAssessorController.cs
```diff
@@ -65,32 +79,60 @@ public async Task<List<RoatpAssessorApplicationSummary>> InProgressApplications(
...
+
+                if (applySequence != null)
+                {
+                    foreach (var section in sequence.Sections)
+                    {
+                        var applySection = applySequence?.Sections?.FirstOrDefault(sec => sec.SectionNo == section.SectionNumber);
+
+                        if (applySequence.NotRequired || applySection?.NotRequired == true)
+                        {
+                            section.Status = "Not Required";
```
David-B-Read, line 129: Can this be replaced with a const?

PR #552, opened 2020-05-14
Apr 1702 resolve timeouts

MichaelYoung1981:


src/SFA.DAS.ApplyService.Data/OrganisationRepository.cs
```diff
@@ -88,18 +83,18 @@ FROM Organisations
             }
         }
 
-        public async Task<Organisation> UpdateOrganisation(Organisation organisation)
+        public async Task UpdateOrganisation(Organisation organisation)
```
martin-hornsby, line 86: Wouldn't it be better if this returns a bool or something to flag that an update has been successful?

Also, now that you don't return the Organisation back, there is the potential that the caller will assume that the Organisation that they asked to be updated will have those values.  Properties such as UpdatedBy and UpdatedAt will be out of sync.

Additionally there are several private methods that are updating the Organisation, hence this could exasperate the issue.

src/SFA.DAS.ApplyService.Data/OrganisationRepository.cs
```diff
@@ -88,18 +83,18 @@ FROM Organisations
...
                 if (connection.State != ConnectionState.Open)
                     await connection.OpenAsync();
 
-                return await UpdateOrganisation(organisation, connection);
+                await UpdateOrganisation(organisation, connection);
             }
         }
 
-        private async Task<Organisation> UpdateOrganisation(Organisation organisation, SqlConnection connection)
+        private async Task UpdateOrganisation(Organisation organisation, SqlConnection connection)
```
martin-hornsby, line 97: This looks like it should be a static method

src/SFA.DAS.ApplyService.Data/OrganisationRepository.cs
```diff
@@ -108,50 +103,37 @@ private async Task<Organisation> UpdateOrganisation(Organisation organisation, S
...
                 "WHERE [Id] = @Id",
                 new {organisation.Id, organisation.Name, organisation.OrganisationType, organisation.OrganisationUkprn, organisation.OrganisationDetails, organisation.UpdatedBy, organisation.RoEPAOApproved, organisation.RoATPApproved});
-
-            var org = await GetOrganisationByName(organisation.Name);
-
-            return org;
         }
 
-        public async Task<Organisation> UpdateOrganisation(Organisation organisation, Guid userId)
+        public async Task UpdateOrganisation(Organisation organisation, Guid userId)
```
martin-hornsby, line 108: same comment as above regarding UpdateOrganisation

PR #554, opened 2020-05-27
Apr 1661 sector employee experience

MarkFCain:


src/SFA.DAS.ApplyService.InternalApi.UnitTests/RoatpAssessorControllerGetSectorDetailsTests.cs
```diff
@@ -0,0 +1,295 @@
...
+        private readonly Guid _applicationId = Guid.NewGuid();
+        private readonly int _sequenceId = RoatpWorkflowSequenceIds.DeliveringApprenticeshipTraining;
+
+        private readonly int _sectionId =
+            RoatpWorkflowSectionIds.DeliveringApprenticeshipTraining.YourSectorsAndEmployees;
+
+        private readonly string _firstPageId = "7610";
+        private readonly string _secondPage = "7611";
+        private readonly string _thirdPage = "7612";
+        private readonly string _fourthPage = "7613";
```
martin-hornsby, line 32: Can these use the constants in RoatpPreambleQuestionBuilder.cs ?

martin-hornsby, line 32: Can these use the constants in RoatpPreambleQuestionBuilder.cs ?

src/SFA.DAS.ApplyService.InternalApi.UnitTests/Services/AssessorLookupServiceTests.cs
```diff
@@ -56,5 +56,16 @@ public void GetTitleForPage_when_invalid_page_returns_null(string invalidPageId)
...
             Assert.That(actualTitle, Is.Null);
         }
+
+        [Test]
+        public void GetSectorIdsForSectorPageId()
+        {
+            var pageId = RoatpWorkflowPageIds.DeliveringApprenticeshipTraining.AgricultureEnvironmentalAndAnimalCare.MostExperiencedEmployee;
+            var result = _assessorLookupService.GetSectorQuestionIdsForSectorPageId(pageId);
+
+            Assert.AreEqual(
```
martin-hornsby, line 66: Can you use CollectionAssert.AreEquivalent here instead?

src/SFA.DAS.ApplyService.InternalApi/Controllers/RoatpAssessorController.cs
```diff
@@ -193,6 +193,107 @@ private static List<int> GetWhatYouWillNeedSectionsForSequence(int sequenceNumbe
         }
 
 
+        [HttpGet("Assessor/Applications/{applicationId}/SectorDetails/{pageId}")]
+        public async Task<SectorDetails> GetSectorDetails(Guid applicationId, string pageId)
```
David-B-Read, line 197: Can this method be broken up a bit?

src/SFA.DAS.ApplyService.InternalApi/Services/AssessorLookupService.cs
```diff
@@ -1,4 +1,6 @@
 ﻿using SFA.DAS.ApplyService.Application.Apply.Roatp;
+using SFA.DAS.ApplyService.Domain.Sectors;
+using static SFA.DAS.ApplyService.Application.Apply.Roatp.RoatpWorkflowPageIds.DeliveringApprenticeshipTraining;
```
martin-hornsby, line 3: Using this static has made a lot of the constants in the case statements look really weird. Do you really need it?

src/SFA.DAS.ApplyService.Application/Apply/Roatp/RoatpPreambleQuestionBuilder.cs
```diff
@@ -317,13 +319,30 @@ public static class DeliveringApprenticeshipTraining
             public const string ProfessionalDevelopmentPolicy = "7700";
             public const string ImproveEmployeeSectorExpertise = "7710";
             public const string ImproveEmployeeKnowledge = "7720";
+            public const string DeliveringTrainingOther = "other";
 
             public static class AgricultureEnvironmentalAndAnimalCare
```
David-B-Read, line 324: Can these be moved into their own file(s)? This RoatpPreambleQuestionBuilder file is becoming a lot more than its original purpose

David-B-Read, line 324: Can these be moved into their own file(s)? This RoatpPreambleQuestionBuilder file is becoming a lot more than its original purpose

src/SFA.DAS.ApplyService.InternalApi/Controllers/RoatpAssessorController.cs
MarkFCain: Yes, I'll break it into private methods with clear naming

src/SFA.DAS.ApplyService.InternalApi.UnitTests/RoatpAssessorControllerGetSectorDetailsTests.cs
MarkFCain: They are just made up numbers, and not significant for the test to work - but I know they appear to be genuine real values.  I'll change them to prevent any confusion

src/SFA.DAS.ApplyService.InternalApi.UnitTests/Services/AssessorLookupServiceTests.cs
MarkFCain: They are not collections.  They're a class which contains a lot of fields that are Ids

MarkFCain: I tried to use the .That and 'AreSame'/'AreEqual' convention to compare, but Nunit wasn't having any of it

src/SFA.DAS.ApplyService.InternalApi/Services/AssessorLookupService.cs
MarkFCain: No, but it makes my stuff REALLY long.  However, I may rejig the class structures to be shorter

src/SFA.DAS.ApplyService.Application/Apply/Roatp/RoatpPreambleQuestionBuilder.cs
MarkFCain: Agree

MarkFCain: I'll move it into a partial.  If any more get out of hand, we can do the same with them

src/SFA.DAS.ApplyService.InternalApi.UnitTests/Services/AssessorLookupServiceTests.cs
martin-hornsby: Fair enough. Marking as resolved

src/SFA.DAS.ApplyService.InternalApi/Services/AssessorLookupService.cs
```diff
@@ -233,6 +233,45 @@ public string GetTitleForPage(string pageId)
             }
         }
 
+        public string GetSectorNameForPage(string pageId)
+        {
+            switch (pageId)
+            {
+                case RoatpWorkflowPageIds.DeliveringApprenticeshipTraining.AgricultureEnvironmentalAndAnimalCare.MostExperiencedEmployee:
+                    return RoatpWorkflowPageIds.DeliveringApprenticeshipTraining.AgricultureEnvironmentalAndAnimalCare.Name;
```
David-B-Read, line 241: Could this be a dictionary, with the page ID as a key?

MarkFCain: I tried this and it's clunky (would need exception handling to deal with code flow) when a pageId is entered that doesn't exist - (this is possible through the API call), so I'm leaving as is.

PR #560, opened 2020-05-29
Apr 1633 assessor moderation tab

MichaelYoung1981:


src/SFA.DAS.ApplyService.Data/AssessorRepository.cs
```diff
@@ -151,6 +151,44 @@ public async Task<int> GetInProgressAssessorApplicationsCount(string userId)
...
+                        $@"SELECT COUNT(1)
+	                      FROM Apply apply
+	                      WHERE (Assessor1ReviewStatus = @approvedReviewStatus OR Assessor2ReviewStatus = @approvedReviewStatus) AND ModerationStatus <> @completedModerationStatus",
+                        new
+                        {
+                            approvedReviewStatus = AssessorReviewStatus.Approved,
+                            completedModerationStatus = ModerationStatus.Complete
+                        }));
+            }
+        }
```
MarkFCain, line 190: I think you need to add 'AND apply.DeletedAt IS NULL' to both of these

src/SFA.DAS.ApplyService.Database/Tables/Apply.sql
```diff
@@ -24,6 +24,7 @@
     [Assessor1ReviewStatus] NVARCHAR(20) NULL, 
     [Assessor2ReviewStatus] NVARCHAR(20) NULL,
     [NotRequiredOverrides] NVARCHAR(MAX) NULL, 
+    [ModerationStatus] NVARCHAR(20) NOT NULL DEFAULT 'New', 
     [OversightStatus] NVARCHAR(20) NOT NULL DEFAULT 'New', 
     [ApplicationDeterminedDate] DATETIME2 NULL
```
MarkFCain, line 29: probably a good idea to get this addition in a little branch of its own and get it into master ASAP so it doesn't break test releases of other branches

src/SFA.DAS.ApplyService.Data/AssessorRepository.cs
martin-hornsby: And possibly an ApplicationStatus check to signify moving from Assessor to Moderation

src/SFA.DAS.ApplyService.Database/Tables/Apply.sql
martin-hornsby: this has been done

PR #571, opened 2020-06-25
QUAL-46 fix async and other compiler warnings

DevenKShah:
I have updated the API project to ignore XML comments warnings as none of the controllers have XML comments. 
There are couple of warnings for unused variable left behind. I did not fix this as there is commented code that uses the variables and I do not have enough understanding to delete it. Also this is not in the scope of the task.

### DIFF
--- a/src/SFA.DAS.ApplyService.Data/OrganisationRepository.cs
+++ b/src/SFA.DAS.ApplyService.Data/OrganisationRepository.cs
@@ -94,9 +94,9 @@ namespace SFA.DAS.ApplyService.Data
             }
         }
 
-        private async Task UpdateOrganisation(Organisation organisation, SqlConnection connection)
+        private Task UpdateOrganisation(Organisation organisation, SqlConnection connection)
         {
-            connection.Execute(
+            return connection.ExecuteAsync(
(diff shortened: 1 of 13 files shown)


src/SFA.DAS.ApplyService.Application/Apply/Review/Return/ReturnRequestHandler.cs
```diff
@@ -36,7 +36,7 @@ public class ReturnRequestHandler : IRequestHandler<ReturnRequest>
             _logger = logger;
         }
 
-        public async Task<Unit> Handle(ReturnRequest request, CancellationToken cancellationToken)
+        public Task<Unit> Handle(ReturnRequest request, CancellationToken cancellationToken)
```
CodeWolf86, line 39: Guessing changing this from a Task<Unit> to just Unit would be troublesome?

dev4valtech: @CodeWolf86  Ideally the whole thing should be deleted or refactored as the comment suggests. I didn't wanted to make too many changes... so just took minimal possible route to get rid of the warning.

PR #581, opened 2020-08-25
Apr 1836 tell us whos in control status

MarkFCain:


src/SFA.DAS.ApplyService.Web.UnitTests/RoatpTaskListWorkflowServiceTests.cs
```diff
@@ -921,6 +921,50 @@ public void Whos_in_control_section_status_shows_as_complete_if_companies_house_
...
+        }
+
+
+        [Test]
+        public void Whos_in_control_section_status_shows_as_next_if_charity_commission_verified_and_not_confirmed_and_who_is_in_control_is_false()
+        {
+            var organisationVerificationStatus = new OrganisationVerificationStatus
+            {
+                VerifiedCompaniesHouse = true,
+                VerifiedCharityCommission = false,
```
martin-hornsby, line 952: think this should be true according to test description

src/SFA.DAS.ApplyService.Web.UnitTests/RoatpTaskListWorkflowServiceTests.cs
```diff
@@ -921,6 +921,50 @@ public void Whos_in_control_section_status_shows_as_complete_if_companies_house_
...
+            status.Should().Be(TaskListSectionStatus.Next);
+        }
+
+
+        [Test]
+        public void Whos_in_control_section_status_shows_as_next_if_charity_commission_verified_and_not_confirmed_and_who_is_in_control_is_false()
+        {
+            var organisationVerificationStatus = new OrganisationVerificationStatus
+            {
+                VerifiedCompaniesHouse = true,
```
martin-hornsby, line 951: should this be true as CompaniesHouseDataConfirmed is false so I think it will ask you to enter CH directors, not Trustees

src/SFA.DAS.ApplyService.Web/Services/RoatpTaskListWorkflowService.cs
```diff
@@ -302,8 +303,14 @@ private string WhosInControlSectionStatus(Guid applicationId, IEnumerable<Applic
...
                 }
 
-                var companiesHouseVerified = organisationVerificationStatus.CompaniesHouseDataConfirmed 
-                                         || organisationVerificationStatus.CompaniesHouseManualEntry;
+
+                if ((organisationVerificationStatus.CompaniesHouseManualEntry || organisationVerificationStatus.CharityCommissionManualEntry) 
+                    && !organisationVerificationStatus.CharityCommissionDataConfirmed
+                    && !organisationVerificationStatus.CompaniesHouseDataConfirmed
+                    && !organisationVerificationStatus.WhosInControlConfirmed)
+                    return TaskListSectionStatus.Next;
```
martin-hornsby, line 311: Can you confirm this says In Progress when you're part way through entering Directors/Trustees

martin-hornsby (changes requested):

src/SFA.DAS.ApplyService.Web/Services/RoatpTaskListWorkflowService.cs
MarkFCain: You can't be part-way through.  The flow is now you enter one and it's confimed, there's no confirm at the end,, so there is no such thing as 'part way'.  If you have entered one, you can finish at that point and you are 'complete'.  The wording 'confirm and save' should have been changed to something like 'continue and save' a long time ago, as we do with management hierarchy.  I can't rmember why this was never done.

src/SFA.DAS.ApplyService.Web.UnitTests/RoatpTaskListWorkflowServiceTests.cs
MarkFCain: fixed

src/SFA.DAS.ApplyService.Web/Services/RoatpTaskListWorkflowService.cs
MarkFCain: That's current, accepted flow btw :+1:

MarkFCain: Yes, in management hierarchy, as you add each one, it's 'confirm and continue', and the list page is just 'continue'..  People in control elements are for each one you enter 'confirm and continue', but the list page also says 'confirm and continue'.  I remember arguing long ago it should say 'continue' to be clearer to the user, but it never got changed.

MarkFCain: This means there's no concept of 'in progress' for these items, they're either done (1 or more) or not done/next (0)

martin-hornsby: sounds good 👍

martin-hornsby (approved):

PR #585, opened 2020-09-09
Apr 1816 fix incorrect paging based on route change

MarkFCain:
This cannot be released without the corresponding 1816 branch in qna being released first


src/SFA.DAS.ApplyService.Web.UnitTests/Controllers/RoatpApplicationPreambleControllerTests.cs
```diff
@@ -45,7 +46,7 @@ public class RoatpApplicationPreambleControllerTests
         private Mock<IApplicationApiClient> _applicationApiClient;
         private Mock<IQnaApiClient> _qnaApiClient;
         private Mock<IUkprnWhitelistValidator> _ukprnWhitelistValidator;
-
+        private Mock<IResetRouteQuestionsService> _resetRoutQuestionsService;
```
martin-hornsby, line 49: minor spelling error here

martin-hornsby (approved):

PR #599, opened 2020-10-19
APR-1721 clarification requested on added

MarkFCain:


martin-hornsby (approved):

src/SFA.DAS.ApplyService.Data/AssessorRepository.cs
```diff
@@ -226,7 +226,7 @@ public async Task<List<ModerationApplicationSummary>> GetApplicationsInModeratio
                             , ModerationStatus
 	                        FROM Apply apply
 	                        INNER JOIN Organisations org ON org.Id = apply.OrganisationId
-	                        WHERE {InModerationApplicationsWhereClause}
```
martin-hornsby, line 229: Can this be changed back to using this where clause (update it if appropriate) as it's also required for the count for the dashboards.

martin-hornsby (changes requested):

src/SFA.DAS.ApplyService.Data/AssessorRepository.cs
MarkFCain: ok, I thought I'd merged right

martin-hornsby (approved):

PR #610, opened 2020-11-09
Apr 1958 refactor application submission

chrisfoster76:


src/SFA.DAS.ApplyService.Web/Controllers/Roatp/RoatpApplicationController.cs
```diff
@@ -198,22 +199,9 @@ private async Task<Guid> StartApplication(Guid signinId)
...
-                    {
-                        await SaveCompaniesHouseInformation(applicationId, applicationDetails);
-                        _logger.LogDebug("Companies House information saved");
-                    }
-
-                    if (applicationDetails.UkrlpLookupDetails.VerifiedbyCharityCommission)
-                    {
-                        await SaveCharityCommissionInformation(applicationId, applicationDetails);
-                        _logger.LogDebug("Save Charity Commission information saved");
-                    }
```
martin-hornsby, line 214: As this conditional logic is now removed; does that mean the new code can handle it correctly still?

src/SFA.DAS.ApplyService.Web/Controllers/Roatp/RoatpApplicationController.cs
```diff
@@ -1094,6 +1082,12 @@ private async Task UpdateQuestionsWithinQnA(Guid applicationId, List<PreambleAns
         {
             if (questions != null)
             {
+                _logger.LogDebug("UpdateQuestionsWithinQnA invoked...");
+
+                var sequenceSections = new List<Tuple<int, int>>();
```
martin-hornsby, line 1087: A Tuple with two items smells like a Dictionary to me, or a KeyValuePair...

src/SFA.DAS.ApplyService.Web/Controllers/Roatp/RoatpApplicationController.cs
martin-hornsby: never mind I see it's done via StorePreambleInformation 😀

martin-hornsby (approved):
After speaking to Chris, this code makes as much of multi-threading as possible and doing things sequentially when concurrency may occur. The result of all this makes the code a bit harder to read but in all honesty doesn't look that bad in the grand scheme of things. Approving

PR #616, opened 2020-11-16
Apr 1799 change ukprn journey

chrisfoster76:


src/SFA.DAS.ApplyService.Web/Views/Shared/_ValidationSummaryPartial.cshtml
```diff
@@ -0,0 +1,33 @@
...
+                            {
+                                <li>
+                                    <a href="#@propertyName">
+                                        <span class="govuk-visually-hidden">Error: </span>@error.ErrorMessage
+                                    </a>
+                                </li>
+                            }
+                        }
+                    }
+                }
```
martin-hornsby, line 29: This looks quite verbose and relies on using ModelExplorer which I think won't capture any custom errors saved to the ModelState. Here's what we used else where:
```
                 @foreach (var item in ViewData.ModelState)
                {
                    foreach (var error in item.Value.Errors)
                    {
                        <li><a href="#@item.Key"><span class="govuk-visually-hidden">Error: </span>@error.ErrorMessage</a></li>
                    }
                }
```

src/SFA.DAS.ApplyService.Web/Controllers/Roatp/RoatpApplicationController.cs
```diff
@@ -517,6 +519,86 @@ public async Task<IActionResult> TaskList(Guid applicationId)
...
+        public IActionResult ChangeUkprn(Guid applicationId)
+        {
+            var model = new ChangeUkprnViewModel { ApplicationId = applicationId };
+
+            return View("~/Views/Roatp/ChangeUkprn.cshtml", model);
+        }
+
+        [HttpPost]
+        [Route("change-ukprn")]
+        public async Task<IActionResult> ChangeUkprn(ChangeUkprnViewModel model)
```
martin-hornsby, line 533: Nothing in this method is async

martin-hornsby (changes requested):

martin-hornsby (approved):

PR #622, opened 2020-11-23
APR-1990 : Two in Twelve Months

martin-hornsby:


src/SFA.DAS.ApplyService.Web/ViewModels/Roatp/TwoInTwelveMonthsViewModel.cs
```diff
@@ -0,0 +1,29 @@
...
+using SFA.DAS.ApplyService.Domain.Apply;
+using System.Collections.Generic;
+
+namespace SFA.DAS.ApplyService.Web.ViewModels.Roatp
+{
+    public class TwoInTwelveMonthsViewModel : IPageViewModel
+    {
+        public bool? HasTwoInTwelveMonths { get; set; }
+
+        public string Title { get { return "Have you had 2 unsuccessful applications in the last 12 months?"; } set { } }
```
MarkFCain, line 11: does this need a set? If not, => will do

src/SFA.DAS.ApplyService.Web/ViewModels/Roatp/TwoInTwelveMonthsViewModel.cs
```diff
@@ -0,0 +1,29 @@
...
+
+        public string GetHelpQuestion { get; set; }
+
+        public bool GetHelpQuerySubmitted { get; set; }
+
+        public string GetHelpErrorMessage { get; set; }
+
+        public List<ValidationErrorDetail> ErrorMessages { get; set; }
+
+        public string GetHelpAction { get { return "TwoInTwelveMonths"; } set { } }
```
MarkFCain, line 27: Does this need a set?

MarkFCain (dismissed):
some minor comments that don't really matter.

src/SFA.DAS.ApplyService.Web/ViewModels/Roatp/TwoInTwelveMonthsViewModel.cs
martin-hornsby: abstract base class so needs an implementation :rage1:

src/SFA.DAS.ApplyService.Web/ViewModels/Roatp/TwoInTwelveMonthsViewModel.cs
martin-hornsby: abstract base class so needs an implementation :rage1:

MarkFCain (approved):

PR #632, opened 2020-12-08
Apr 2051 audit gateway outcomes

chrisfoster76:


src/SFA.DAS.ApplyService.Application/Apply/Gateway/UpdateGatewayPageAnswerRequestHandler.cs
```diff
@@ -0,0 +1,76 @@
...
+            {
+                answer = CreateNewGatewayPageAnswer(request.ApplicationId, request.PageId);
+                _auditService.AuditInsert(answer);
+            }
+            else
+            {
+                _auditService.AuditUpdate(answer);
+            }
+
+            _auditService.AuditUpdate(application);
```
MarkFCain, line 40: can you move this adjacent to the .UpdateApplication call, just easier to follow

src/SFA.DAS.ApplyService.Application/Apply/Gateway/UpdateGatewayPageAnswerRequestHandler.cs
```diff
@@ -0,0 +1,76 @@
...
+
+        public async Task<Unit> Handle(UpdateGatewayPageAnswerRequest request, CancellationToken cancellationToken)
+        {
+            _auditService.StartTracking(UserAction.UpdateGatewayPageOutcome, request.UserId, request.UserName);
+
+            var application = await _applyRepository.GetApplication(request.ApplicationId);
+            var answer = await _applyRepository.GetGatewayPageAnswer(request.ApplicationId, request.PageId);
+            var isNew = answer == null;
+
+            if (answer == null)
```
MarkFCain, line 30: change to if (isNew), a bit easier to read

MarkFCain (dismissed):

src/SFA.DAS.ApplyService.Application/Apply/Gateway/UpdateGatewayPageAnswerRequestHandler.cs
chrisfoster76: done

src/SFA.DAS.ApplyService.Application/Apply/Gateway/UpdateGatewayPageAnswerRequestHandler.cs
chrisfoster76: done

src/SFA.DAS.ApplyService.Application.UnitTests/Handlers/Gateway/UpdateGatewayPageAnswerRequestHandlerTests.cs
```diff
@@ -0,0 +1,146 @@
...
+            _autoFixture = new Fixture();
+            _repository = new Mock<IApplyRepository>();
+            _auditService = new Mock<IAuditService>();
+
+            _applicationId = Guid.NewGuid();
+            _pageId = _autoFixture.Create<string>();
+            _userId = _autoFixture.Create<string>();
+            _userName = _autoFixture.Create<string>();
+            _status = _autoFixture.Create<string>();
+            _comments = _autoFixture.Create<string>();
```
martin-hornsby, line 44: These are fine, although I'd personally have used realistic values so that they have more meaning when observing the tests

src/SFA.DAS.ApplyService.Application.UnitTests/Services/AuditServiceTests.cs
```diff
@@ -0,0 +1,222 @@
...
+using SFA.DAS.ApplyService.Application.Interfaces;
+using SFA.DAS.ApplyService.Application.Services;
+using SFA.DAS.ApplyService.Domain.Audit;
+
+namespace SFA.DAS.ApplyService.Application.UnitTests.Services
+{
+    [TestFixture]
+    public class AuditServiceTests
+    {
+        private AuditServiceTestsFixture _fixture;
```
martin-hornsby, line 19: I'm not fully clued up with AutoFixture but a lot of these tests looks like they could be covered with Moq which is used throughout this code base

src/SFA.DAS.ApplyService.Application/Apply/IApplyRepository.cs
```diff
@@ -23,7 +24,10 @@ public interface IApplyRepository
         Task<string> GetGatewayPageStatus(Guid applicationId, string pageId);
         Task<string> GetGatewayPageComments(Guid applicationId, string pageId);
 
-        Task SubmitGatewayPageAnswer(Guid applicationId, string pageId, string userId, string userName, string status, string comments);
+        Task InsertGatewayPageAnswer(GatewayPageAnswer pageAnswer, string userId, string userName);
+        Task UpdateGatewayPageAnswer(GatewayPageAnswer pageAnswer, string userId, string userName);
+        Task UpdateApplication(Domain.Entities.Apply application);
```
martin-hornsby, line 29: Would it be useful if these return a bool so the caller knows if the insert/update was successful

src/SFA.DAS.ApplyService.Database/SFA.DAS.ApplyService.Database.refactorlog
```diff
@@ -0,0 +1,10 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<Operations Version="1.0" xmlns="http://schemas.microsoft.com/sqlserver/dac/Serialization/2012/02">
+  <Operation Name="Rename Refactor" Key="bc1e5a0b-b386-4245-9d4f-05f10d59b0e4" ChangeDateTime="12/07/2020 15:32:02">
+    <Property Name="ElementName" Value="[dbo].[Audit].[OriginalState]" />
+    <Property Name="ElementType" Value="SqlSimpleColumn" />
+    <Property Name="ParentElementName" Value="[dbo].[Audit]" />
+    <Property Name="ParentElementType" Value="SqlTable" />
+    <Property Name="NewName" Value="InitialState" />
+  </Operation>
+</Operations>
```
martin-hornsby, line 10: can we get rid of this file?

src/SFA.DAS.ApplyService.Application/Services/StateService.cs
```diff
@@ -0,0 +1,33 @@
...
+namespace SFA.DAS.ApplyService.Application.Services
+{
+    public class StateService : IStateService
+    {
+        public Dictionary<string, object> GetState(object item)
+        {
+            var result = new Dictionary<string, object>();
+            var targetType = item.GetType();
+
+            foreach (var property in targetType.GetProperties(BindingFlags.Instance | BindingFlags.Public | BindingFlags.NonPublic))
```
martin-hornsby, line 15: Just to check...  _BindingFlags.NonPublic_ is fine here  (i.e. we want to know the state of anything private which would be normally hidden for reason)

src/SFA.DAS.ApplyService.Application/Services/DiffService.cs
```diff
@@ -0,0 +1,78 @@
...
+                    {
+                        PropertyName = item.Key,
+                        InitialValue = initialValue,
+                        UpdatedValue = updatedValue
+                    });
+                }
+            }
+
+            return result;
+        }
```
martin-hornsby, line 76: This is looking a tad chunky. I was wondering if refactoring down into several private methods would be of benefit?

martin-hornsby (dismissed):

src/SFA.DAS.ApplyService.Database/SFA.DAS.ApplyService.Database.refactorlog
chrisfoster76: yep

src/SFA.DAS.ApplyService.Application/Services/StateService.cs
chrisfoster76: I think _NonPublic_ may not actually be required, so I will see if it can come out. I think there was a specific need for it before, when dealing with private fields within rich domain classes, but this won't be the case here.

martin-hornsby (approved):

PR #639, opened 2020-12-18
Apr 2041 gw clarification template

MarkFCain:


src/SFA.DAS.ApplyService.Application.UnitTests/Handlers/Gateway/UpdateGatewayPageAnswerRequestHandlerTests.cs
```diff
@@ -86,41 +89,44 @@ public async Task Handle_Updates_Existing_GatewayPageAnswer()
             ), _userId, _userName));
         }
 
-        [Test]
-        public async Task Handle_Insert_Is_Subject_To_Audit()
+        [TestCase(null)]
```
martin-hornsby, line 92: do you need to worry about string.Emtpy / "" ?

martin-hornsby (approved):

martin-hornsby (approved):

PR #641, opened 2020-12-22
APR-2033 : Application Withdrawal

martin-lush:


src/SFA.DAS.ApplyService.InternalApi/Controllers/GatewayReviewController.cs
```diff
@@ -51,11 +52,24 @@ public async Task EvaluateGateway(Guid applicationId, [FromBody] EvaluateGateway
         {
             await _mediator.Send(new EvaluateGatewayRequest(applicationId, request.IsGatewayApproved, request.EvaluatedBy));
         }
+
+        [HttpPost("GatewayReview/{applicationId}/Withdraw")]
+        public async Task<bool> WithdrawApplication(Guid applicationId, [FromBody] GatewayWithdrawApplicationRequest request)
```
MarkFCain, line 57: Does this need to be a FromBody? Doesn't look like it.  Have a look and leave as is, if it does.

src/SFA.DAS.ApplyService.Web/Controllers/Roatp/RoatpApplicationController.cs
```diff
@@ -105,7 +105,9 @@ public async Task<IActionResult> Applications()
 
             var signinId = await _userService.GetSignInId();
             var applications = await _apiClient.GetApplications(signinId, false);
-            applications = applications.Where(app => app.ApplicationStatus != ApplicationStatus.Rejected && app.ApplicationStatus != ApplicationStatus.Cancelled).ToList();
+
+            var statusFilter = new[] { ApplicationStatus.Rejected, ApplicationStatus.Cancelled, ApplicationStatus.Withdrawn };
```
MarkFCain, line 109: leave for now, but if we're reusing this, or it's possibly going to be more complex, we need a static service that returns this list

MarkFCain (dismissed):
no real changes, approved

src/SFA.DAS.ApplyService.InternalApi/Controllers/GatewayReviewController.cs
martin-hornsby: it contains comments & user info which would be too long via query string

src/SFA.DAS.ApplyService.Web/Controllers/Roatp/RoatpApplicationController.cs
martin-hornsby: agreed. This is going to get expanded on soon

MarkFCain (dismissed):

MarkFCain (approved):

PR #642, opened 2020-12-23
Apr 2045 gw clarification upload template

MarkFCain:


src/SFA.DAS.ApplyService.Data/ApplyRepository.cs
```diff
@@ -212,6 +212,24 @@ public async Task<bool> UpdateGatewayReviewStatusAndComment(Guid applicationId,
             return await Task.FromResult(true);
         }
 
+        public async Task<bool> UpdateGatewayApplyData(Guid applicationId, ApplyData applyData, string userId, string userName)
```
martin-lush, line 215: Possibly a bit dubious as there's an UpdateApplyData method. Maybe we need to think about having a common update ApplyData method that can take in a variety of options. Or we split into a separate GatewayRepository class so the lines aren't as muddied. Raise as tech debt?

martin-lush (approved):

PR #649, opened 2021-01-07
Apr 2079 governance report download

chrisfoster76:


src/SFA.DAS.ApplyService.Application/Apply/IApplyRepository.cs
```diff
@@ -79,6 +79,8 @@ public interface IApplyRepository
 
         Task<bool> UpdateOversightReviewStatus(Guid applicationId, string oversightStatus, string userId, string userName);
         Task<ApplicationOversightDetails> GetOversightDetails(Guid applicationId);
+        Task<List<ApplicationOversightDownloadDetails>> GetOversightsForDownload(DateTime dateFrom, DateTime dateTo);
```
martin-lush, line 82: just for consideration, potentially putting this on line 79 so it's next to the other ones

src/SFA.DAS.ApplyService.Web/Controllers/Roatp/RoatpApplicationController.cs
```diff
@@ -1482,6 +1493,33 @@ private async Task<List<TabularData>> GetPeopleInControlDetails(Guid application
...
+
+            var answers = page.PageOfAnswers.First().Answers;
+            var i = 0;
+
+            foreach (var answer in answers)
+            {
+                if (i > 0) result.Append(", ");
+                result.Append(answer.Value);
+                i++;
+            }
```
martin-lush, line 1518: Are we able to use string.Join here instead or will it give incorrect results?

martin-lush (approved):

src/SFA.DAS.ApplyService.Web/Controllers/Roatp/RoatpApplicationController.cs
chrisfoster76: Done

martin-lush (dismissed):

MarkFCain (approved):

PR #653, opened 2021-01-14
APR-2069 Retrieving review statuses from database for new outcome col…

ConorTill:
…umn in application oversight details


MarkFCain (dismissed):

src/SFA.DAS.ApplyService.Data/ApplyRepository.cs
```diff
@@ -991,6 +991,132 @@ public async Task<Organisation> GetOrganisationForApplication(Guid applicationId
             }
         }
 
+        public async Task<List<ApplicationOversightDetails>> GetOversightsPending()
```
chrisfoster76, line 994: @ConorTill You've added these repo methods (back?) into this class, but there don't seem to be any usages. Could you check, and remove if not needed.

chrisfoster76 (approved):

PR #661, opened 2021-01-25
APR-2063 : Withdrawal email notification

martin-lush:


MarkFCain (dismissed):
There might be a case for putting the withdraw processing into its own service so RoatpGatewayController doesn't need to have dependencies on IConfigurationService and IWithdrawApplicationConfirmationEmailService 

We can discuss if you wish to think about it.  But it is fine otherwise.

gdbarnes (approved):

PR #669, opened 2021-01-29
APR-1609 code and coverage

MarkFCain:


src/SFA.DAS.ApplyService.Web/Controllers/Roatp/RoatpApplicationController.cs
```diff
@@ -508,7 +510,7 @@ public async Task<IActionResult> SaveAnswers(PageViewModel vm, Guid applicationI
                 }
                 else
                 {
-                    return RedirectToAction("TaskList", new { applicationId = applicationId });
+                    return RedirectToAction("TaskList", new { applicationId, sequenceId });
```
chrisfoster76, line 513: These changes, along with the TaskList change below look like they'll just cause a double-redirection of the user? Is there not some other way to modify the qs here?

src/SFA.DAS.ApplyService.Web/Controllers/Roatp/RoatpApplicationController.cs
```diff
@@ -526,7 +540,9 @@ public async Task<IActionResult> TaskList(Guid applicationId)
             }
 
             var viewModel = await _taskListOrchestrator.GetTaskListViewModel(applicationId, User.GetUserId());
-            return View("~/Views/Roatp/TaskList.cshtml", viewModel);
+            var view = View("~/Views/Roatp/TaskList.cshtml", viewModel);
```
chrisfoster76, line 543: You could just return this without the assignment

chrisfoster76 (dismissed):
I just wondered if we could lose the double-redirect

src/SFA.DAS.ApplyService.Web/Views/Roatp/TaskList.cshtml
```diff
@@ -22,8 +21,8 @@
             <li>
                 <h2 class="app-task-list__section">
                     <span class="app-task-list__section-number">@Html.Raw(sequence.Id). </span>
-                    @Html.Raw(sequence.Description)
-                </h2>
+                    <a id="Sequence_@sequence.Id">@Html.Raw(sequence.Description)</a>
```
martin-lush, line 24: Please revert this and just put an `id="Sequence_@sequence.Id"` as now all descriptions are hyperlinks (which they shouldn't be. Also they are not styled correctly ;)

martin-lush (changes requested):

src/SFA.DAS.ApplyService.Web/Controllers/Roatp/RoatpApplicationController.cs
MarkFCain: every other way involved trying to directly inject the anchor into the url, and if you do this in MVC, in escapes the hashtag character and doesn't work

src/SFA.DAS.ApplyService.Web/Controllers/Roatp/RoatpApplicationController.cs
```diff
@@ -517,7 +518,16 @@ public async Task<IActionResult> SaveAnswers(PageViewModel vm, Guid applicationI
...
+        public IActionResult TaskList(Guid applicationId, int? sequenceId)
+        {
+            return sequenceId != null ? 
+                RedirectToAction("TaskListRefresh","RoatpApplication", new {applicationId},"Sequence_" + sequenceId) : 
+                RedirectToAction("TaskListRefresh","RoatpApplication", new { applicationId });
+        }
+
+        [Route("application-tasklist")]
+        [HttpGet]
+        public async Task<IActionResult> TaskListRefresh(Guid applicationId)
```
martin-lush, line 530: Do you still need TaskListRefresh and TaskList based upon recent changes?

martin-lush (changes requested):

src/SFA.DAS.ApplyService.Web/Controllers/Roatp/RoatpApplicationController.cs
```diff
@@ -487,7 +486,9 @@ public async Task<IActionResult> SaveAnswers(PageViewModel vm, Guid applicationI
             
             if (page == null)
             {
-                return RedirectToAction("TaskList", new { applicationId = applicationId });
+                return sequenceId != null ?
```
martin-lush, line 489: line 435: var sequenceId = int.Parse(vm.SequenceId); // TODO: SequenceId should be an int, not a string

Either that above is a now a bug as it can be null. Or the sequenceId  != null check here will always be true

martin-lush (changes requested):

martin-lush (approved):

martin-lush (approved):

PR #676, opened 2021-02-03
Apr 2107 oversight gw reject view

chrisfoster76:
Target 2161 for diff


MarkFCain (approved):

src/SFA.DAS.ApplyService.Application/Apply/Oversight/RecordOversightOutcomeHandler.cs
```diff
@@ -32,43 +32,102 @@ public async Task<bool> Handle(RecordOversightOutcomeCommand request, Cancellati
...
+            var application = await GetApplicationOrThrow(request.ApplicationId);
+
+            var (oversightReview, isNew) = await GetExistingOrNewOversightReview(request.ApplicationId);
+
+            ApplyChanges(oversightReview, request, isNew);
+            await SaveChanges(oversightReview, application, isNew);
+            
+            await _auditService.Save();
+
+            return true;
```
martin-lush, line 46: Would a better check be asserting that at least 1 row was affected on SaveChanges?

martin-lush (approved):

PR #680, opened 2021-02-08
Apr 2195 tech debt move apply service repo interfaces

chrisfoster76:


martin-lush (approved):
All good. As discussed, I think there's dead code in 

- EmailTemplateName.cs
- CreateAccountHandler.cs
- CreateOrganisationHandler.cs

MarkFCain (approved):

PR #681, opened 2021-02-09
Apr 2173 add withdrawn application to overall outcome table

ConorTill:


src/SFA.DAS.ApplyService.Application/Apply/Gateway/ApplicationActions/WithdrawApplicationHandler.cs
```diff
@@ -8,19 +12,39 @@ namespace SFA.DAS.ApplyService.Application.Apply.Gateway.ApplicationActions
...
+                ApplicationId = request.ApplicationId,
+                Status = OversightReviewStatus.Withdrawn,
+                InternalComments = request.Comments,
+                UserId = request.UserId,
+                UserName = request.UserName
+            };
+
+            var result = await _applyRepository.WithdrawApplication(request.ApplicationId, request.Comments, request.UserId, request.UserName);
+
+            _auditService.StartTracking(UserAction.RecordOversightOutcome, request.UserId, request.UserName);
```
chrisfoster76, line 42: The user action here should be the name of the use case, eg "WithdrawApplication", which you'll need to add to the enum.

chrisfoster76 (changes requested):
One minor comment. In addition, I don't see a unit test asserting that an oversight review was created, which would be worth having.

chrisfoster76 (dismissed):

chrisfoster76 (approved):

PR #685, opened 2021-02-10
Apr 2174 removed apps in oversight 2

chrisfoster76:


src/SFA.DAS.ApplyService.Application/Behaviours/UnitOfWorkBehaviour.cs
```diff
@@ -0,0 +1,34 @@
...
+                await _unitOfWork.Commit();
+            }
+
+            return result;
+        }
+
+        private bool IsCommand(TRequest request)
+        {
+            return request.GetType().Name.EndsWith("Command");
+        }
```
martin-lush, line 32: This looks a tad brittle to me. Would it be too much effort to have a base class or check for an interface?

chrisfoster76: I was going for a convention-based approach to identify commands and their handlers.

martin-lush (dismissed):

martin-lush (approved):

PR #686, opened 2021-02-11
Apr 2181 apply prg

chrisfoster76:


MarkFCain (approved):

src/SFA.DAS.ApplyService.Web/Extensions/RestoreModelStateFilter.cs
```diff
@@ -0,0 +1,23 @@
...
+    public class RestoreModelStateFilter : ActionFilterAttribute
+    {
+        public override void OnActionExecuting(ActionExecutingContext filterContext)
+        {
+            var tempDataFactory = filterContext.HttpContext.RequestServices.GetRequiredService<ITempDataDictionaryFactory>();
+            var tempData = tempDataFactory.GetTempData(filterContext.HttpContext);
+            var serializableModelState = tempData.Get<SerializableModelStateDictionary>();
+            var modelState = serializableModelState?.ToModelState();
+
+            filterContext.ModelState.Merge(modelState);
```
martin-lush, line 20: filterContext is not checking it's a Controller so you could theoretically put this attribute on anything and I think the middleware layer will do weird unexpected stuff

src/SFA.DAS.ApplyService.Web/Extensions/SerializableModelStateDictionaryExtensions.cs
```diff
@@ -0,0 +1,24 @@
...
+
+namespace SFA.DAS.ApplyService.Web.Extensions
+{
+    public static class SerializableModelStateDictionaryExtensions
+    {
+        public static ModelStateDictionary ToModelState(this SerializableModelStateDictionary serializableModelState)
+        {
+            var modelState = new ModelStateDictionary();
+
+            foreach (var data in serializableModelState.Data)
```
martin-lush, line 11: No null check (not that I'd expect it to be otherwise but it is a public method)

src/SFA.DAS.ApplyService.Web/Extensions/TempDataDictionaryExtensions.cs
```diff
@@ -0,0 +1,18 @@
+﻿using Microsoft.AspNetCore.Mvc.ViewFeatures;
+using Newtonsoft.Json;
+
+namespace SFA.DAS.ApplyService.Web.Extensions
+{
+    public static class TempDataDictionaryExtensions
+    {
+        public static T Get<T>(this ITempDataDictionary tempData) where T : class
```
martin-lush, line 8: I don't think we use structs, but isn't this something we should have? You could then check for ISerializable too

martin-lush (changes requested):

src/SFA.DAS.ApplyService.Web/Extensions/SerializableModelStateDictionaryExtensions.cs
chrisfoster76: Null check on what, sorry?

src/SFA.DAS.ApplyService.Web/Extensions/TempDataDictionaryExtensions.cs
chrisfoster76: Sorry, I don't follow - what change are you asking for?

src/SFA.DAS.ApplyService.Web/Extensions/SerializableModelStateDictionaryExtensions.cs
martin-lush: serializableModelState.Data could be null

src/SFA.DAS.ApplyService.Web/Extensions/TempDataDictionaryExtensions.cs
martin-lush: T is constrained to class. It is possible to have structs that implement ISerializable

Hence I'd expect T to be new() and do a check that it is a class type or implements ISerializable

If it's overkill don't worry about it

martin-lush: actually ignore me. If we make this too flexible the code checks will become hideous and we're allowing other devs to have types that aren't really allowed in viewmodels

PR #704, opened 2021-03-09
Sending notification email when application is rejected

ConorTill:


src/SFA.DAS.ApplyService.Application.UnitTests/Handlers/UpdateGatewayReviewStatusAndCommentCommandHandlerTests/UpdateGatewayReviewStatusAndCommentCommandHandlerTests.cs
```diff
@@ -75,5 +80,21 @@ public async Task UpdateGatewayReviewStatusAndComment_when_rejected_creates_over
 
             _oversightReviewRepository.Verify(x => x.Add(It.Is<OversightReview>(r => r.ApplicationId == _applicationId && r.Status == OversightReviewStatus.Rejected)));
         }
+
+        [Test]
+        public async Task Handler_sends_updated_email()
```
chrisfoster76, line 85: Is it worth adding another test (or set of TestCases) to verify that in cases other than Reject, _no_ email is sent?

src/SFA.DAS.ApplyService.Application/Apply/Gateway/UpdateGatewayReviewStatusAndCommentCommandHandler.cs
```diff
@@ -59,6 +63,8 @@ public async Task<Unit> Handle(UpdateGatewayReviewStatusAndCommentCommand reques
                 _auditService.AuditInsert(oversightReview);
                 _oversightReviewRepository.Add(oversightReview);
                 _auditService.Save();
+
+                await _applicationUpdatedEmailService.SendEmail(request.ApplicationId);
```
chrisfoster76, line 67: As it stands, this email will be sent and subsequently the database transaction will be committed. In the event of a db connection failure or similar, we'd send the email without having made a state change. It might be worth forcing the commit immediately before this line, so that this can never happen.

To do so, inject the `IUnitOfWork` dependency into the handler, and call its `Commit` method immediately before `SendEmail`.

Happy to discuss.

chrisfoster76 (changes requested):
Looks okay, just a couple of minor points.

src/SFA.DAS.ApplyService.Application.UnitTests/Handlers/UpdateGatewayReviewStatusAndCommentCommandHandlerTests/UpdateGatewayReviewStatusAndCommentCommandHandlerTests.cs
```diff
@@ -96,5 +98,27 @@ public async Task Handler_sends_updated_email()
...
+            var gatewayReviewComment = "Some comment";
+            var gatewayReviewExternalComment = "Some external comment";
+
+            _applyRepository.Setup(x => x.UpdateGatewayReviewStatusAndComment(_applicationId, It.IsAny<ApplyData>(), gatewayReviewStatus, _userId, _userName)).ReturnsAsync(true);
+
+            var request = new UpdateGatewayReviewStatusAndCommentCommand(_applicationId, gatewayReviewStatus, gatewayReviewComment, gatewayReviewExternalComment, _userId, _userName);
+
+            await _handler.Handle(request, CancellationToken.None);
+
+            _applicationUpdatedEmailService.Verify(x => x.SendEmail(It.Is<Guid>(id => id == _applicationId)), Times.Never);
```
chrisfoster76, line 121: You could use It.IsAny<Guid>() here instead, but not a major issue

chrisfoster76 (approved):

PR #709, opened 2021-03-16
UpdateApplicationStatus now requires username and updates updatedby a…

ConorTill:
…nd updatedat. Changed EnterNewUkprn to conform.


src/SFA.DAS.ApplyService.Web/Controllers/Roatp/RoatpApplicationController.cs
```diff
@@ -580,8 +580,8 @@ public async Task<IActionResult> EnterNewUkprn(EnterNewUkprnViewModel model)
                     UKPRN = ukprn,
                     UkrlpLookupDetails = ukrlpLookupResults.Results.FirstOrDefault()
                 };
-
-                await _applicationApiClient.UpdateApplicationStatus(model.ApplicationId, ApplicationStatus.Cancelled);
+                var userName  = User.Identity.Name;
```
martin-lush, line 583: It's wrong to assume this. We don't use names but Guids.

I think it should be User.GetUserId()

martin-lush (changes requested):

martin-lush (approved):

PR #715, opened 2021-03-26
Apr 2138 apply on roatp remove shutter

MarkFCain:


src/SFA.DAS.ApplyService.Web/Controllers/Roatp/RoatpApplicationPreambleController.cs
```diff
@@ -528,12 +531,16 @@ public async Task<IActionResult> VerifyOrganisationDetails()
 
             if (ProviderEligibleToChangeRoute(roatpRegisterStatus))
             {
+                if (applicationDetails?.RoatpRegisterStatus?.ProviderTypeId != null)
+                {
+                    return await TermsAndConditions(new SelectApplicationRouteViewModel { ApplicationRouteId = applicationDetails.RoatpRegisterStatus.ProviderTypeId.Value });
```
martin-lush, line 536: Shouldn't this be a RedirectToAction so the URL is correct?  It's showing `..\choose-provider-route` instead of `..\terms-conditions-making-application'

martin-lush (changes requested):

src/SFA.DAS.ApplyService.Web/Controllers/Roatp/RoatpApplicationPreambleController.cs
MarkFCain: good spot, fixed

martin-lush (approved):

PR #763, opened 2021-05-06
Apr 2481 missing authorisation

chrisfoster76:


src/SFA.DAS.ApplyService.Web/Authorisation/AuthorizationHandler.cs
```diff
@@ -0,0 +1,83 @@
...
+            _apiClient = apiClient;
+            _userService = userService;
+            _logger = logger;
+        }
+
+        public async Task HandleAsync(AuthorizationHandlerContext context)
+        {
+            _logger.LogInformation("AccessApplicationRequirementHandler invoked");
+
+            var requestedApplicationId = _httpContextAccessor.HttpContext.Request.Query["ApplicationId"].ToString();
```
martin-lush, line 34: I can't remember but does this return an empty string or null if the QueryString isn't there?

src/SFA.DAS.ApplyService.Web/Authorisation/AuthorizationHandler.cs
```diff
@@ -0,0 +1,83 @@
...
+                if (requirement is AccessApplicationRequirement)
+                {
+                    if (application != null && application.ApplicationId == applicationId)
+                    {
+                        context.Succeed(requirement);
+                        _logger.LogInformation($"Requirement {requirement.GetType().Name} met");
+                    }
+                    else
+                    {
+                        _logger.LogInformation($"Requirement {requirement.GetType().Name} not met");
```
martin-lush, line 62: Are we okay logging these in Kibana or would they be more suitable as Debug?

src/SFA.DAS.ApplyService.Web/Authorisation/AuthorizationHandler.cs
```diff
@@ -0,0 +1,83 @@
...
+                var signInId = await _userService.GetSignInId();
+                application = await _apiClient.GetApplicationByUserId(applicationId, signInId);
+            }
+            else
+            {
+                _logger.LogInformation("Unable to determine ApplicationId parameter");
+            }
+
+            var pendingRequirements = context.PendingRequirements.ToList();
+            foreach (var requirement in pendingRequirements)
```
martin-lush, line 49: If there are multiple requirements. One could success and one could fail and the logic below to me looks like it would still wave them through

src/SFA.DAS.ApplyService.Web/Authorisation/AuthorizationHandler.cs
martin-lush: Unless it's supposed to be a If - else if statement then we're fine 🤷

src/SFA.DAS.ApplyService.Web/Authorisation/AuthorizationHandler.cs
martin-lush: Unless it's supposed to be a If - else if statement then we're fine 🤷

martin-lush (approved):

martin-lush (approved):

PR #768, opened 2021-05-12
Apr 2495 change of route fix

MarkFCain:


martin-lush (approved):

src/SFA.DAS.ApplyService.Web.UnitTests/Controllers/RoatpApplicationPreambleControllerTests.cs
```diff
@@ -1599,7 +1599,7 @@ public void Provider_changing_route_mid_application_currently_active_on_the_regi
 
             model.ApplicationId.Should().Be(applicationId);
             model.ApplicationRouteId.Should().Be(applyApplicationRouteId);
-            model.ApplicationRoutes.Should().NotContain(route => route.Id == existingProviderStatus.ProviderTypeId);
+            model.ApplicationRoutes.Should().Contain(route => route.Id == existingProviderStatus.ProviderTypeId);
```
chrisfoster76, line 1602: It looks like this asserts only that their current route is offered, which is not what the test name suggests.

chrisfoster76 (dismissed):
Approved with minor comment on test

martin-lush (approved):

PR #777, opened 2021-05-21
APR-2518 : Duplicate Request Header key hotfix

martin-lush:


MarkFCain (approved):

martin-lush:
Fix does not address the issue in the ticket but it does resolve some tech debt. I'll raise a new ticket for that

PR #795, opened 2021-06-07
APR-2537 : Implemented Managed Identity for Apply Internal API

martin-lush:


src/SFA.DAS.ApplyService.InternalApi/Authentication/AzureAdScopeClaimTransformation.cs
```diff
@@ -0,0 +1,35 @@
...
+using System.Threading.Tasks;
+
+namespace SFA.DAS.ApplyService.InternalApi.Authentication
+{
+    public class AzureAdScopeClaimTransformation : IClaimsTransformation
+    {
+        public Task<ClaimsPrincipal> TransformAsync(ClaimsPrincipal principal)
+        {
+            var scopeClaims = principal.FindAll(Constants.ScopeClaimType).ToList();
+            if (scopeClaims.Count != 1 || !scopeClaims[0].Value.Contains(' '))
```
MarkFCain, line 15: I think this needs to ben scopeClaims.Count>1 and you might need to cope with scopeClaims.Count==0

wouldn't scopeClaims.First().Value be more readable maybe?

martin-lush: Not sure. This was copied directly from a common package (would have used it but not compatible with Apply, sadly)

MarkFCain (dismissed):

MarkFCain (approved):

PR #798, opened 2021-06-14
APR-2548_code complate

MarkFCain:


martin-lush (approved):

src/SFA.DAS.ApplyService.Web/Controllers/Roatp/RoatpOverallOutcomeController.cs
```diff
@@ -64,6 +63,9 @@ public async Task<IActionResult> ProcessApplicationStatus(Guid applicationId)
                 case ApplicationStatus.Withdrawn:
                     return View("~/Views/Roatp/ApplicationWithdrawn.cshtml", model);
                 case ApplicationStatus.Removed:
+                    var oversightStatus = await _overallOutcomeService.GetOversightReviewStatus(applicationId);
```
chrisfoster76, line 66: You've added a new method `GetOversightReviewStatus` and used it here, but why not just use `GetOversightReview` instead? That would save adding new methods, and the performance difference would be negligible.

Also it seems odd to get the review using the api client, but the status via the overall outcome service, no?

src/SFA.DAS.ApplyService.Web/Services/OverallOutcomeService.cs
```diff
@@ -123,7 +126,11 @@ public async Task<ApplicationSummaryWithModeratorDetailsViewModel> BuildApplicat
             return model;
         }
 
-
+        public async Task<OversightReviewStatus?> GetOversightReviewStatus(Guid applicationId)
```
chrisfoster76, line 129: Why would this be nullable? Isn't that what the `None` entry is for?

src/SFA.DAS.ApplyService.Web/Controllers/Roatp/RoatpOverallOutcomeController.cs
MarkFCain: yeah, I'm mixing and matching.  Will conform

src/SFA.DAS.ApplyService.Web/Services/OverallOutcomeService.cs
MarkFCain: taking this out

martin-lush (dismissed):

martin-lush (approved):

PR #805, opened 2021-06-18
Apr 2398 supporting 500k outcome

MarkFCain:


src/SFA.DAS.ApplyService.Web/ViewModels/Roatp/ApplicationSummaryViewModel.cs
```diff
@@ -19,6 +19,11 @@ public class ApplicationSummaryViewModel
         public string GatewayReviewStatus { get; set; }
         public string ModerationStatus { get; set; }
         public string OversightInProgressExternalComments { get; set; }
+
+        public int? SubcontractingLimit { get; set; }
+
+        public string SubcontractingLimitFormatted => !string.IsNullOrEmpty(SubcontractingLimit?.ToString()) ?SubcontractingLimit?.ToString("N0") : null;
```
martin-lush, line 25: you only need: SubcontractingLimit?.ToString("N0")  due to null coalesce

martin-lush (changes requested):

martin-lush (dismissed):

PR #809, opened 2021-06-25
Apr 2566 unsuccessful gw pass oversight overturned

MarkFCain:


src/SFA.DAS.ApplyService.Web/Services/OverallOutcomeService.cs
```diff
@@ -2,6 +2,7 @@
 using System.Collections.Generic;
 using System.Linq;
 using System.Threading.Tasks;
+using NPOI.SS.Formula.Functions;
```
martin-lush, line 5: I don't think this should be here

src/SFA.DAS.ApplyService.Web/Services/OverallOutcomeService.cs
```diff
@@ -97,38 +98,48 @@ public ApplicationSummaryViewModel BuildApplicationSummaryViewModel(Apply applic
...
-                    applicationUnsuccessfulModerationFail = true;
+                    if (oversightReview.ModerationApproved.HasValue && oversightReview.ModerationApproved == true)
+                    {
+                        moderationFailedAndApproved = true;
+                    }
+
+                    if (oversightReview.ModerationApproved.HasValue && oversightReview.ModerationApproved == false)
+                    {
+                        moderationFailedAndOverturned = true;
+                    }
```
martin-lush, line 124: This took me a few moments to figure out what was happening. How about this to cover both cases:

`if (oversightReview.ModerationApproved.HasValue)
 {
        moderationFailedAndApproved = oversightReview.ModerationApproved;
 }`

martin-lush (changes requested):

src/SFA.DAS.ApplyService.Web/Services/OverallOutcomeService.cs
martin-lush, line 124: Took a few moments to see what was going on here. Personally I would make it a:  _if - else if_ statement

src/SFA.DAS.ApplyService.Web/Services/OverallOutcomeService.cs
```diff
@@ -97,38 +98,48 @@ public ApplicationSummaryViewModel BuildApplicationSummaryViewModel(Apply applic
...
+            var moderationPassedAndApproved = false;
+            var gatewayPassOverturnedToFail = false;
+            var moderationFailedAndOverturned = false;
             if (application?.GatewayReviewStatus == GatewayAnswerStatus.Pass)
             {
-                if (application?.ModerationStatus != null
-                    && application?.ModerationStatus == ModerationStatus.Fail
-                    && oversightReview.ModerationApproved.HasValue
-                    && oversightReview.ModerationApproved == true)
+                if (application?.ModerationStatus != null && application?.ModerationStatus == ModerationStatus.Fail)
```
martin-lush, line 114: can remove `application?.ModerationStatus != null` as the null coalesce will do that

src/SFA.DAS.ApplyService.Web/Services/OverallOutcomeService.cs
```diff
@@ -97,38 +98,48 @@ public ApplicationSummaryViewModel BuildApplicationSummaryViewModel(Apply applic
...
+                        moderationFailedAndApproved = true;
+                    }
+
+                    if (oversightReview.ModerationApproved.HasValue && oversightReview.ModerationApproved == false)
+                    {
+                        moderationFailedAndOverturned = true;
+                    }
                 }
 
                 if (application?.ModerationStatus != null
```
martin-lush, line 127: same here as above can remove `application?.ModerationStatus != null` as the null coalesce will do that

martin-lush (approved):

src/SFA.DAS.ApplyService.Web/Services/OverallOutcomeService.cs
MarkFCain: will remove

PR #892, opened 2021-10-05
APR-2738 application queries updated to exclude appeals

MarkFCain:


src/SFA.DAS.ApplyService.Data/Queries/OversightReviewQueries.cs
```diff
@@ -45,7 +45,9 @@ public async Task<PendingOversightReviews> GetPendingOversightReviews(string sea
                               FROM Apply apply
 	                      INNER JOIN Organisations org ON org.Id = apply.OrganisationId
                          LEFT JOIN OversightReview r ON r.ApplicationId = apply.ApplicationId
```
martin-lush, line 47: Is it me or isn't this line doing anything? Should you also check  'AND r.ApplicationId IS NULL'

src/SFA.DAS.ApplyService.Data/Queries/OversightReviewQueries.cs
```diff
@@ -45,7 +45,9 @@ public async Task<PendingOversightReviews> GetPendingOversightReviews(string sea
...
+                         LEFT JOIN Appeal Appeal on apply.ApplicationId = Appeal.ApplicationId
 	                      WHERE apply.DeletedAt IS NULL
+                          AND Appeal.ApplicationId IS NULL
```
martin-lush, line 50: The oversight review check below is filtering on Status, can you match it to do the same check: `Appeal.Status IS NULL` or do `r.ApplicationId IS NULL`

src/SFA.DAS.ApplyService.Data/Queries/OversightReviewQueries.cs
```diff
@@ -94,7 +96,9 @@ public async Task<CompletedOversightReviews> GetCompletedOversightReviews(string
                               FROM Apply apply
 	                      INNER JOIN Organisations org ON org.Id = apply.OrganisationId
                           INNER JOIN OversightReview r ON r.ApplicationId = apply.ApplicationId
+                          LEFT JOIN Appeal Appeal on apply.ApplicationId = Appeal.ApplicationId
 	                      WHERE apply.DeletedAt IS NULL
+                          AND Appeal.ApplicationId IS NULL
```
martin-lush, line 101: Match to the comment above if doing changes

martin-lush (changes requested):

src/SFA.DAS.ApplyService.Data/Queries/OversightReviewQueries.cs
MarkFCain: changed to Appeal.Status

src/SFA.DAS.ApplyService.Data/Queries/OversightReviewQueries.cs
MarkFCain: changed to Appeal.Status

martin-lush (approved):

PR #895, opened 2021-10-07
Apr 2731 appeal successful feedback

MAli99:


src/SFA.DAS.ApplyService.Web/Controllers/Roatp/RoatpAppealsController.cs
```diff
@@ -228,6 +229,19 @@ public async Task<IActionResult> AppealSuccessful(Guid applicationId)
                 SubcontractingLimit = application?.ApplyData?.GatewayReviewDetails?.SubcontractingLimit
             };
 
+            var isGatewayFail = application?.GatewayReviewStatus == GatewayReviewStatus.Fail;
+            if (isGatewayFail)
+            {
+                switch (appeal?.Status)
```
martin-lush, line 235: appeal won't be null at this point (due to line 210) so you can remove the null coalesce

PR #921, opened 2021-12-08
APR-2868 scripts to populate EmailTemplates added

MarkFCain:


src/SFA.DAS.ApplyService.Database/Script.PostDeployment1.sql
```diff
@@ -9,3 +9,28 @@ Post-Deployment Script Template
                SELECT * FROM [$(TableName)]					
 --------------------------------------------------------------------------------------
 */
+
+-- SET UP EmailTemplates
+IF NOT EXISTS (SELECT * FROM EmailTemplates WHERE TemplateName = N'RoATPRequestInvitationToReapply')
```
DevenKShah, line 14: This needs to be in its own file and use MERGE statement. The file should be then included here as shown on line #6

MarkFCain: ok I'll make that change, but we usually only have done that for transitory script changes (eg setup that needs to propagate just once through all release environments).  This will be a permanent setup, so I thought inline would be better.

DevenKShah (approved):

PR #925, opened 2021-12-16
APR-2859-added legend for radio buttons

raviudari01:


DevenKShah (dismissed):

src/SFA.DAS.ApplyService.Web/Views/Users/ExistingAccount.cshtml
```diff
@@ -23,6 +22,9 @@
                     <span class="govuk-error-message" asp-validation-for="FirstTimeSignin"></span>
```
MarkFCain, line 22: I think this needs to go below 'the heading'.  So move to line 28, and move everything else down a line?

raviudari01: Done

DevenKShah (dismissed):

jamesmking (approved):

DevenKShah (approved):

PR #934, opened 2022-01-07
APR-2909-PRG pattern in RoatpAppealsController without modelstate

raviudari01:


src/SFA.DAS.ApplyService.Web/Controllers/Roatp/RoatpAppealsController.cs
```diff
@@ -140,7 +136,7 @@ public async Task<IActionResult> GroundsOfAppeal(GroundsOfAppealViewModel model)
 
             if (!ModelState.IsValid)
             {
-                return RedirectToAction("GroundsOfAppeal", new { model.ApplicationId, model.AppealOnPolicyOrProcesses, model.AppealOnEvidenceSubmitted });
+                return View("~/Views/Appeals/GroundsOfAppeal.cshtml", model);
```
DevenKShah, line 139: You need to populate appeals file list in the model here, see the GET method. Refactor the model building logic in the GET method so it can be reused here.

DevenKShah (changes requested):

src/SFA.DAS.ApplyService.Web/Controllers/Roatp/RoatpAppealsController.cs
raviudari01: Refactored the code, but calling model build is not required in the post as it is retaining the removed fail

DevenKShah (approved):

PR #935, opened 2022-01-07
Apr 2893 prg ukprn entry

MarkFCain:


src/SFA.DAS.ApplyService.Web/Controllers/Roatp/RoatpApplicationPreambleController.cs
```diff
@@ -142,6 +143,23 @@ public IActionResult EnterApplicationUkprn(string ukprn)
             {
                 model.UKPRN = ukprn;
             }
+           
+           var ukprnTempData = TempData[_ukprn] as string;
+
+           if (!string.IsNullOrEmpty(ukprnTempData))
+           {
+               model.UKPRN = ukprnTempData;
```
DevenKShah, line 151: There are two ways UKPRN is being passed in the method now. I think we should stick to either passing it via model/query param or via tempdata. I prefer former as it was already being used.

src/SFA.DAS.ApplyService.Web.UnitTests/Controllers/RoatpApplicationPreambleControllerTests.cs
```diff
@@ -274,12 +278,12 @@ public void Before_each_test()
...
-        [TestCase("1234567")]
-        [TestCase("123456789")]
-        [TestCase("1234567A")]
-        public void Validation_error_is_triggered_if_UKPRN_not_in_correct_format(string ukprn)
+        [TestCase("", _validationTypeMissing)]
+        [TestCase(null, _validationTypeMissing)]
+        [TestCase("1234567",_validationTypeInvalid)]
+        [TestCase("123456789", _validationTypeInvalid)]
+        [TestCase("1234567A", _validationTypeInvalid)]
+        public void Validation_error_is_triggered_if_UKPRN_not_in_correct_format(string ukprn, string validationMessageType)
```
DevenKShah, line 286: The test names are not according to recommendations as described [here](https://docs.microsoft.com/en-us/dotnet/core/testing/unit-testing-best-practices#naming-your-tests)

Rename suggestion SearchByUkprn_InvalidInput_RedirectsToEnterUkrpnWithRespectiveValidationMessage

DevenKShah (approved):
Looks fab, just couple of suggestions.

PR #936, opened 2022-01-11
APR-2894 already an employer provider code and coverage complete

MarkFCain:


src/SFA.DAS.ApplyService.Web/Controllers/Roatp/RoatpApplicationPreambleController.cs
```diff
@@ -712,28 +727,14 @@ public async Task<IActionResult> ChangeProviderRoute(ChangeProviderRouteViewMode
...
-
-                var modelErrors = ModelState.Values.SelectMany(v => v.Errors);
-                foreach (var modelError in modelErrors)
-                {
-                    model.ErrorMessages.Add(new ValidationErrorDetail
-                    {
-                        Field = "ChangeApplicationRoute",
-                        ErrorMessage = modelError.ErrorMessage
-                    });
+                    TempData[_employerProviderValidationMessage] = modelErrors.ErrorMessage;
```
DevenKShah, line 734: We should be returning view here as the model state is invalid, isn't it?

src/SFA.DAS.ApplyService.Web/Views/Roatp/ProviderAlreadyOnRegister.cshtml
```diff
@@ -21,7 +21,12 @@
                     </h2>
                     <div class="govuk-error-summary__body">
                         <ul class="govuk-list govuk-error-summary__list">
-                            <li><a sfa-validation-for="ChangeApplicationRoute"></a></li>
+                            @foreach (var error in Model.ErrorMessages)
```
DevenKShah, line 24: What warranted this change?

src/SFA.DAS.ApplyService.Web/Controllers/Roatp/RoatpApplicationPreambleController.cs
MarkFCain: that's what causes the post to get put into the browser history, so we are avoiding that in all the PRG stories.

src/SFA.DAS.ApplyService.Web/Views/Roatp/ProviderAlreadyOnRegister.cshtml
MarkFCain: Validation stopped working, in that the top message wasn't being written.  I tried various methods to get it working using the current setup, but none worked, so went with this pattern (which is also present in various preamble pages)

src/SFA.DAS.ApplyService.Web/Controllers/Roatp/RoatpApplicationPreambleController.cs
MarkFCain: It follows the pattern established in APR-2893 of writing the message to tempData and redirecting back to the GET

PR #938, opened 2022-01-12
APR-2895-choose provider flows-PRG

raviudari01:


src/SFA.DAS.ApplyService.Web/Controllers/Roatp/RoatpApplicationPreambleController.cs
```diff
@@ -599,7 +616,6 @@ public async Task<IActionResult> ChangeApplicationProviderRoute(Guid application
             return View("~/Views/Roatp/SelectApplicationRoute.cshtml", model);
         }
 
-        [HttpPost]
```
MarkFCain, line 602: As we discussed, Ravi, I'm not a fan of just losing a VERB on a controller method, i'd typically prefer to add a new method with the missing verb , but in the case of 'UpdateApplicationProviderRoute', it's doing an awful lot already and it would be somewhat clunky to wrap all that in a private method for the POST and the GET, so in this case, I think it's acceptable.

MarkFCain (dismissed):

src/SFA.DAS.ApplyService.Web.UnitTests/Controllers/RoatpApplicationPreambleControllerTests.cs
```diff
@@ -1810,10 +1810,7 @@ public void Provider_changing_route_to_employer_is_directed_to_conditions_of_acc
 
             var redirectResult = result as RedirectToActionResult;
             redirectResult.Should().NotBeNull();
-            redirectResult.ActionName.Should().Be("ConditionsOfAcceptance");
-            redirectResult.ControllerName.Should().BeNull();
-
-            _qnaApiClient.VerifyAll();
+            redirectResult.ActionName.Should().Be("UpdateApplicationProviderRoute");
```
DevenKShah, line 1813: How come the redirection has changed? Is this change in behaviour?

src/SFA.DAS.ApplyService.Web/Controllers/Roatp/RoatpApplicationPreambleController.cs
```diff
@@ -71,6 +71,7 @@ public class RoatpApplicationPreambleController : RoatpApplyControllerBase
         }
 
         [Route("conditions-of-acceptance")]
+        [HttpGet]
         public async Task<IActionResult> ConditionsOfAcceptance(SelectApplicationRouteViewModel routeViewModel)
```
DevenKShah, line 75: Remove async Task from the signature as there are no awaited task in the method.

src/SFA.DAS.ApplyService.Web/Controllers/Roatp/RoatpApplicationPreambleController.cs
DevenKShah: I am interested in knowing why this is taken  off?

src/SFA.DAS.ApplyService.Web/Controllers/Roatp/RoatpApplicationPreambleController.cs
```diff
@@ -412,7 +429,7 @@ public async Task<IActionResult> ProcessRoute(SelectApplicationRouteViewModel mo
             }
             else
             {
-                return await ConditionsOfAcceptance(new SelectApplicationRouteViewModel { ApplicationRouteId = model.ApplicationRouteId });
+                return RedirectToAction("ConditionsOfAcceptance", model);
```
DevenKShah, line 432: Redirection has changed?

src/SFA.DAS.ApplyService.Web/Controllers/Roatp/RoatpApplicationPreambleController.cs
```diff
@@ -399,8 +399,25 @@ public async Task<IActionResult> SelectApplicationRoute()
             return View("~/Views/Roatp/SelectApplicationRoute.cshtml", model);
         }
 
+        [HttpPost]
         public async Task<IActionResult> ProcessRoute(SelectApplicationRouteViewModel model)
         {
+            if (!ModelState.IsValid)
+            {
+                model.ApplicationRoutes = await GetApplicationRoutes();
```
DevenKShah, line 407: Why is the standard way of reporting validation errors not being used here?

DevenKShah (commented):
Few questions

src/SFA.DAS.ApplyService.Web.UnitTests/Controllers/RoatpApplicationPreambleControllerTests.cs
raviudari01: when implemented PRG, "UpdateApplicationProviderRoute" is the first redirection

src/SFA.DAS.ApplyService.Web/Controllers/Roatp/RoatpApplicationPreambleController.cs
raviudari01: Done

src/SFA.DAS.ApplyService.Web/Controllers/Roatp/RoatpApplicationPreambleController.cs
raviudari01: View requires ApplicationRoutes, which is not there in the model until we fill it

DevenKShah (approved):
Looks good.

PR #944, opened 2022-01-19
APR-2930 Table structure for data enhancements

DevenKShah:


src/SFA.DAS.ApplyService.Database/Tables/OrganisationSectorExperts.sql
```diff
@@ -0,0 +1,29 @@
...
+	[SectorTrainingExperienceDuration] NVARCHAR(30) NULL,
+	[SectorTrainingExperienceDetails] NVARCHAR(MAX) NULL,
+	[IsQualifiedForSector] BIT NOT NULL,
+	[QualificationDetails] NVARCHAR(MAX) NULL,
+	[IsApprovedByAwardingBodies] BIT NOT NULL, 
+	[AwardingBodyNames] NVARCHAR(MAX) NULL, 
+	[HasSectorOrTradeBodyMembership] BIT NOT NULL, 
+	[SectorOrTradeBodyNames] NVARCHAR(MAX) NULL, 
+	[TypeOfApprenticeshipDelivered] NVARCHAR(50) NULL, 
+	[ExperienceInTrainingApprentices] NVARCHAR(100) NULL,
```
MarkFCain, line 24: ExperienceInTrainingApprentices needs to be 255 characters

MarkFCain (approved):

src/SFA.DAS.ApplyService.Database/Tables/OrganisationSectorExperts.sql
DevenKShah: Following are the expected input for this, so I guess 100 should be sufficient 
image

src/SFA.DAS.ApplyService.Database/Tables/OrganisationAddresses.sql
```diff
@@ -0,0 +1,14 @@
+﻿CREATE TABLE [dbo].[OrganisationAddresses]
+(
+	[Id] INT NOT NULL IDENTITY(1,1),
+	[OrganisationId] UNIQUEIDENTIFIER NOT NULL,
+	[AddressType] [TINYINT] NOT NULL,
+	[AddressLine1] NVARCHAR(MAX) NULL,
+	[AddressLine2] NVARCHAR(MAX) NULL,
+	[AddressLine3] NVARCHAR(MAX) NULL,
```
raviudari01, line 8: Do AddressLine1,2,3 needs to be Max size ?

raviudari01 (approved):

src/SFA.DAS.ApplyService.Database/Tables/OrganisationAddresses.sql
DevenKShah: Address is not user input, we derive it from external API so it really does not matter.

PR #947, opened 2022-01-26
Apr 2914 admin download legal address das-apply-service

raviudari01:


DevenKShah (commented):
Has this requirement been met `And the column header should read : Legal address`?

DevenKShah (dismissed):

DevenKShah:
> Has this requirement been met `And the column header should read : Legal address`?

I see this is handled in the other PR

DevenKShah (approved):

PR #962, opened 2022-03-18
Apr 2852 section15 fix

MarkFCain:


src/SFA.DAS.ApplyService.Web/Services/ResetQnaDetailsService.cs
```diff
@@ -0,0 +1,47 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Threading.Tasks;
+using SFA.DAS.ApplyService.Application.Apply.Roatp;
+using SFA.DAS.ApplyService.Web.Infrastructure;
+
+namespace SFA.DAS.ApplyService.Web.Services
+{
+    public class ResetQnaDetailsService : IResetQnaDetailsService
```
DevenKShah, line 9: Rename ResetCompleteFlagService

src/SFA.DAS.ApplyService.Web/Services/ResetQnaDetailsService.cs
```diff
@@ -0,0 +1,47 @@
...
+    {
+        private readonly IQnaApiClient _qnaApiClient;
+
+        public ResetQnaDetailsService(IQnaApiClient qnaApiClient)
+        {
+            _qnaApiClient = qnaApiClient;
+        }
+
+        public async Task ResetPagesComplete(Guid applicationId, int sequenceNo, int sectionNo, string pageId)
```
DevenKShah, line 18: Remove this method.

src/SFA.DAS.ApplyService.Web/Services/ResetQnaDetailsService.cs
```diff
@@ -0,0 +1,47 @@
...
+        {
+            await ResetSection1_5(applicationId, sequenceNo, sectionNo, pageId);
+        }
+
+        private async Task ResetSection1_5(Guid applicationId, int sequenceNo, int sectionNo, string pageId)
```
DevenKShah, line 23: Make this method public and get take only one parameter i.e. applicationId

src/SFA.DAS.ApplyService.Web/Infrastructure/QnaApiClient.cs
```diff
@@ -405,6 +405,10 @@ public async Task DeleteFile(Guid applicationId, Guid sectionId, string pageId,
             }
         }
 
+        public async Task ResetSectionPagesIncomplete(Guid applicationId, int sequenceNo, int sectionNo, List<string> pageIdsExcluded)
```
DevenKShah, line 408: Rename ResetCompleteFlag

src/SFA.DAS.ApplyService.Web/Infrastructure/QnaApiClient.cs
```diff
@@ -405,6 +405,10 @@ public async Task DeleteFile(Guid applicationId, Guid sectionId, string pageId,
...
+        {
+            await _httpClient.PostAsJsonAsync($"/applications/{applicationId}/sequences/{sequenceNo}/sections/{sectionNo}/reset-complete", pageIdsExcluded);
```
DevenKShah, line 410: rename endpoint to reset-completeflag

src/SFA.DAS.ApplyService.Web/Services/ResetQnaDetailsService.cs
MarkFCain: We only want the reset to happen when a specific sequence/section/page is saved for the application.  We need all the parameters, otherwise it's move all the logic in this service back into the controller, which is cluttering the controller.

MarkFCain: I changed this to use minimum number of parameters, which is applicationId and pageId

src/SFA.DAS.ApplyService.Web/Services/ResetCompleteFlagService.cs
```diff
@@ -0,0 +1,38 @@
...
+            const string entryPageForSupportingRoute = RoatpWorkflowPageIds.ExperienceAndAccreditations.SubcontractorDeclaration;
+
+            var pageIdsToExcludeFromCompleteReset = new List<string>
+            {
+                entryPageForMainRoute,
+                entryPageForEmployerRoute,
+                entryPageForSupportingRoute
+            };
+
+            if (pageIdsToExcludeFromCompleteReset.Contains(pageId))
```
DevenKShah, line 31: If not in exclude right? Or consider renaming the variable to have `included`?

DevenKShah (approved):
Couple of minor things, but otherwise looks ok.

PR #963, opened 2022-03-23
APR-2985-Gateway-checks-for-prefixed-charities-bug-fix

raviudari01:


MarkFCain (changes requested):
this is an oc

src/SFA.DAS.ApplyService.InternalApi/Services/GatewayApiChecksService.cs
```diff
@@ -110,7 +110,7 @@ private async Task LookupCharityCommissionDetails(ApplyGatewayDetails applyGatew
                 var message =
                     $"Unable to parse charity registration number from charity number in ukrlp: '{charityNumberFromUkrlp}'";
                 _logger.LogError(message);
```
MarkFCain, line 112: this is no longer an error, but an occasional, but expected condition.  Can you change .LogError to .LogInformation

DevenKShah: Perhaps change the message as well. To something like `Charity registration number cannot be parsed into a number, skipping getting details from charity commission for {charityNumber}`

MarkFCain (dismissed):

src/SFA.DAS.ApplyService.InternalApi/Services/GatewayApiChecksService.cs
raviudari01: Done

DevenKShah (approved):

PR #978, opened 2022-10-31
CSP-325-DotCore6Migration-apply

raviudari01:


src/SFA.DAS.ApplyService.Data/OrganisationRepository.cs
```diff
@@ -111,18 +111,6 @@ public async Task UpdateOrganisation(Organisation organisation, Guid userId)
             }
         }
 
-        public async Task UpdateOrganisation(Guid organisationId, Guid userId)
```
MarkFCain, line 114: I wouldn't make any code changes, can you put this back?

MarkFCain (changes requested):
Can you reset the index name back to what it was, and put back in the method OrganisationRepository.UpdateOrganisation

src/SFA.DAS.ApplyService.Data/SFA.DAS.ApplyService.Data.csproj
```diff
@@ -13,7 +13,7 @@
 
     <ItemGroup>
       <PackageReference Include="Dapper">
-        <Version>1.50.5</Version>
```
DevenKShah, line 16: Move version on the same line

MarkFCain (dismissed):

src/SFA.DAS.ApplyService.Web/Views/Home/Error.cshtml
```diff
@@ -10,32 +8,9 @@
 <main class="govuk-main-wrapper govuk-main-wrapper--l" id="main-content" role="main">
     <div class="govuk-grid-row">
         <div class="govuk-grid-column-two-thirds">
-            @if (HostingEnvironment.IsDevelopment())
```
DevenKShah, line 13: We removed this with agreement that in the startup we will configure to go to this page only if it is on hosted environment, we still need to see `DeveloperExceptionPage` when debugging. I have not found any config changes in this regards?

DevenKShah (commented):
Developer exception page needs to be enabled.

src/SFA.DAS.ApplyService.Web/Startup.cs
```diff
@@ -36,20 +33,25 @@
...
 {
+    [ExcludeFromCodeCoverage]
     public class Startup
     {
         private readonly IConfiguration _configuration;
         private readonly ILogger<Startup> _logger;
-        private readonly IHostingEnvironment _hostingEnvironment;
-        private readonly IHostingEnvironment _env;
+        private readonly IWebHostEnvironment _hostingEnvironment;
+        private readonly IWebHostEnvironment _env;
```
DevenKShah, line 49: Do we need two variables for the same thing?

DevenKShah (dismissed):

MarkFCain (dismissed):

HussainsGitHub (approved):

DevenKShah (approved):

PR #1010, opened 2024-11-22
CSP-1787 add trustee exempted list service

MarkFCain:


src/SFA.DAS.ApplyService.Web/Controllers/Roatp/RoatpWhosInControlApplicationController.cs
```diff
@@ -219,6 +223,15 @@ public async Task<IActionResult> ConfirmTrustees(Guid applicationId)
         {
             var trusteesData = await _tabularDataRepository.GetTabularDataAnswer(applicationId, RoatpWorkflowQuestionTags.CharityCommissionTrustees);
 
+            var qnaApplicationData = await _qnaApiClient.GetApplicationData(applicationId);
+            var ukprn = qnaApplicationData.GetValue(RoatpWorkflowQuestionTags.UKPRN)?.Value<string>();
```
DevenKShah, line 227: Ukprn is already being passed in to this method, is there any other purpose for this call?

src/SFA.DAS.ApplyService.Web/Services/TrusteeExemptionService.cs
```diff
@@ -0,0 +1,30 @@
...
+    private readonly IConfigurationService _configurationService;
+
+    public TrusteeExemptionService(IConfigurationService configurationService)
+    {
+        _configurationService = configurationService;
+    }
+
+    public async Task<bool> IsProviderExempted(string ukprn)
+    {
+        var config = await _configurationService.GetConfig();
```
DevenKShah, line 18: Can we not use IOptions to load this value at the start up so that config is loaded only once per application lifecycle?

src/SFA.DAS.ApplyService.Web/Services/TrusteeExemptionService.cs
```diff
@@ -0,0 +1,30 @@
...
+        var exemptionListing = config.ProvidersExemptedFromHavingTrustees;
+        var exemptions = exemptionListing.Split(',');
+
+        var isPresent = exemptions.Any(x => x == ukprn);
+        return isPresent;
+    }
+}
+
+public interface ITrusteeExemptionService
```
DevenKShah, line 27: Can the interface be moved on the top of the file or to its own file?

src/SFA.DAS.ApplyService.Web/Services/RoatpOrganisationVerificationService.cs
```diff
@@ -91,6 +94,15 @@ private static bool WhosInControlStarted(JObject qnaApplicationData)
             return soleTraderOrPartnership == RoatpOrganisationTypes.SoleTrader || soleTraderOrPartnership == RoatpOrganisationTypes.Partnership;
         }
 
+        private async Task<bool> CharityCommissionDataExempted(JObject qnaApplicationData)
```
DevenKShah, line 97: Have you checked if the consumer has the ukprn handy and which can be passed in as a param here?

PeterDavidWilliams (approved):

src/SFA.DAS.ApplyService.Web.UnitTests/Controllers/RoatpWhosInControlApplicationControllerTests.cs
```diff
@@ -511,6 +523,66 @@ public void Confirm_trustees_presents_list_of_trustees()
             trusteesCount.Should().Be(2);
         }
 
+
+        [Test]
+        public void Confirm_trustees_exempted_returns_to_tasklist()
```
DevenKShah, line 528: Test is not named to show its purpose.... `Confirm_NoTrusteesAndProviderExempted_ReturnsToTaskList`

src/SFA.DAS.ApplyService.Web.UnitTests/Controllers/RoatpWhosInControlApplicationControllerTests.cs
```diff
@@ -511,6 +523,66 @@ public void Confirm_trustees_presents_list_of_trustees()
...
+            var qnaApplicationData = new JObject
+            {
+                [RoatpWorkflowQuestionTags.UKPRN] = ukprn
+            };
+
+            _qnaClient.Setup(x => x.GetApplicationData(It.IsAny<Guid>())).ReturnsAsync(qnaApplicationData);
+
+            _qnaClient.Setup(x => x.GetAnswerByTag(It.IsAny<Guid>(), RoatpWorkflowQuestionTags.UkrlpVerificationCompany, It.IsAny<string>())).ReturnsAsync(verifiedCompanyAnswer);
+
+            var result = _controller.ConfirmTrustees(Guid.NewGuid()).GetAwaiter().GetResult();
```
DevenKShah, line 550: Use async test instead

src/SFA.DAS.ApplyService.Web/Services/RoatpOrganisationVerificationService.cs
MarkFCain: It doesn't

src/SFA.DAS.ApplyService.Web/Controllers/Roatp/RoatpWhosInControlApplicationController.cs
MarkFCain: I don't understand this comment, ukprn is not being passed into this method. only applicationId

src/SFA.DAS.ApplyService.Web/Services/TrusteeExemptionService.cs
MarkFCain: moved to its own file

src/SFA.DAS.ApplyService.Web.UnitTests/Controllers/RoatpWhosInControlApplicationControllerTests.cs
MarkFCain: ok

src/SFA.DAS.ApplyService.Web.UnitTests/Controllers/RoatpWhosInControlApplicationControllerTests.cs
MarkFCain: ok

src/SFA.DAS.ApplyService.Web/Services/TrusteeExemptionService.cs
MarkFCain: _configureService is already cached. Isn't that the same?

DevenKShah (approved):

PR #1012, opened 2024-12-31
DASD-9588 - Convert das-apply-service from classic to YAML

furqanagwan:


azure-pipelines.yml
```diff
@@ -1,128 +1,160 @@
...
+- name: Deploy
+  value: $[or(eq(variables['Build.SourceBranch'], 'refs/heads/master'), eq(variables['Build.Reason'], 'Manual'), eq(variables['Build.Reason'], 'PullRequest'))]
 
 resources:
   repositories:
+  - repository: self
   - repository: das-platform-building-blocks
     type: github
     name: SkillsFundingAgency/das-platform-building-blocks
     ref: refs/tags/2.1.28
```
calumrees99, line 31: Please bump this to current latest version

calumrees99 (changes requested):
We should break this down into multiple jobs, das-recruits looks like a good example of a pipeline that deploys multiple apps:

https://github.com/SkillsFundingAgency/das-recruit/blob/master/pipeline-templates/job/deploy.yml

azure-pipelines.yml
furqanagwan: Bumped to 2.2.17

